PR #17, opened 2023-08-17
Feat/aave borrowing strat kdlv2

lnist:
Outstanding work (some also needed for borrow kandel):
- Sanity-check PoC code
- 	Remove compounding
- 	Remove todos
- 	Test it all
- 	Update natspec where applicable
- 	Rename/delete geometricKandel?


lnist:
This PoC has the advantage of handling pending, but other than that it is surpassed by the mgvv2 kandel-strat.

PR #18, opened 2023-08-17
Feat/peter/use ticks

peterMangrove:


lnist:
Replaced with https://github.com/mangrovedao/mangrove-strats/pull/33 to track work

PR #28, opened 2023-08-28
feat: use mgvConfig for addresses

peterMangrove:
This PR uses the mgvConfig introduced in mangrove-core. 
It adds the dependancy addresses paths to the configuration, so that mangrove-strats work out of the box.


.github/workflows/node.js.yml
```diff
@@ -120,13 +120,13 @@ jobs:
...
+    - mangrove-strats
 
     runs-on: ubuntu-latest
 
     steps:
     - name: Decide whether the needed jobs succeeded or failed
       uses: re-actors/alls-green@release/v1
       with:
-        allowed-skips: file-guard, mangrove-core
+        allowed-skips: file-guard, mangrove-starts
```
espendk, line 131:
```suggestion
        allowed-skips: file-guard, mangrove-strats
```

espendk (changes requested):
lgtm except for small typ

espendk (approved):

PR #33, opened 2023-09-01
Upgrade strats to new core

lnist:
Changes since Audit March 2023 (3bff09efba82a6d55d19eeb807654833339785f1), see AUDIT_TRAIL.md

TODO:

- [x] Upgrade to newer core (OLKey, LogPrice)
- [x] Remove pivot concept
- [x] Remove compounding from Kandel
- [x] Same event fixes as in https://github.com/mangrovedao/mangrove-core/pull/460
- [x] fix boundary-tests
- [x] Consider tickScale>1
- [x] Consider precision loss
- [x] other fixmes
- [x] contracts too big
- [x] Rename spread to stepSize (Check with Jean first)
- [x] cleanup - uptake new core with a lot of renames
- [x] Audit guide
- [x] Use prerelease core package released from core's develop branch.
- [ ] Merge with develop
- [ ] nice-to-have before audit:
  - [x] Consider deleting/moving unaudited files like Mango, OfferMaker to toy or delete
  - [x] Consider moving everything out of "strategies" folder (no, since we also have core folder in core, and toy_strategies)
  - [x] Consider packing arguments for populate(Chunk)FromOffset for L2 (no, postponing for later
  - [x] Consider if precision can be increased in tests by changing to TickOffset for assertions
  - [x] Consider: ask/bid distribution -> into a struct with ask/bid/reservedAsks/reservedBids, latter without gives. See https://github.com/mangrovedao/mangrove-strats/pull/64 for PoC - too gas costly and size increasing.
  - [ ] Measure Gas (including gasreq - see fixme's)
  - [x] Update readme - it is not strats-specific


src/strategies/offer_maker/market_making/mango/Mango.sol
espendk, line 1: Is Mango worth keeping around?

test/toy_strategies/Amplifier.t.sol
```diff
@@ -122,17 +122,19 @@ contract AmplifierTest is StratTest {
     });
   }
 
-  function takeOffer(uint makerGivesAmount, uint makerWantsAmount, IERC20 makerWantsToken, uint offerId)
+  function takeOffer(uint makerWantsAmount, IERC20 makerWantsToken, uint offerId)
     public
     returns (uint takerGot, uint takerGave, uint bounty)
   {
+    Tick tick = mgv.offers($(weth), $(makerWantsToken), offerId).tick();
     // try to snipe one of the offers (using the separate taker account)
```
espendk, line 130: I noticed "snipe" still occurs in many comments

src/strategies/offer_maker/market_making/mango/Mango.sol
lnist: We have a lot of toy contracts. I've updated them to the new API and made tests green, but that is it. No thorough review of those.

lnist: We could have deleted them, but now the work is done, so I'm ok with keeping.

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -272,26 +254,22 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
   /// However one would then need to pay attention to the following fact:
   /// if `order.inbound_tkn` is not pushed to reserve during `makerExecute`, in the posthook of this offer execution, the `order.inbound_tkn` balance of this contract would then contain
```
jkrivine, line 255: couple of `order.in/outbound_tkn` to be replaced here

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
```diff
@@ -154,31 +152,24 @@ abstract contract GeometricKandel is CoreKandel {
...
   ///@notice This function is not payable, use `populate` to fund along with populate.
   ///@param distribution the distribution of base and quote for Kandel indices.
-  ///@param pivotIds the pivot to be used for the offer.
   ///@param firstAskIndex the (inclusive) index after which offer should be an ask.
-  function populateChunk(Distribution calldata distribution, uint[] calldata pivotIds, uint firstAskIndex)
-    external
-    onlyAdmin
-  {
-    populateChunk(distribution, pivotIds, firstAskIndex, params.gasreq, params.gasprice);
+  function populateChunk(Distribution calldata distribution, uint firstAskIndex) external onlyAdmin {
```
jkrivine, line 171: Since mangrove now accepts (price, volume) offers, wouldn't it be simpler to have distribution as:
* logPrice dist
* volume dist
Geometric Kandel would only require giving minLogPrice and would compute the rest of the dist of prices onchain (saves calldata size)

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -79,8 +80,8 @@ contract MumbaiMangroveFullTestnetDeployer is Deployer {
       mgv: mgv,
       gaspriceOverride: 140, // this overrides Mangrove's gasprice for the computation of market's density
       reader: reader,
-      tkn1: dai,
-      tkn2: usdc,
+      //FIXME: what tickscale?
+      market: Market({tkn0: dai, tkn1: usdc, tickScale: 1}),
```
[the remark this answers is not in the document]
lnist: No - this is the definition that is used to activate the market, so there is no tickScale before this.

src/strategies/MangroveOrder.sol
```diff
@@ -37,69 +38,60 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
 
   ///@inheritdoc IOrderLogic
   ///@dev We also allow Mangrove to call this so that it can part of an offer logic.
-  function setExpiry(IERC20 outbound_tkn, IERC20 inbound_tkn, uint offerId, uint date)
-    public
-    mgvOrOwner(outbound_tkn, inbound_tkn, offerId)
-  {
-    expiring[outbound_tkn][inbound_tkn][offerId] = date;
-    emit SetExpiry(address(outbound_tkn), address(inbound_tkn), offerId, date);
+  function setExpiry(bytes32 olKeyHash, uint offerId, uint date) public mgvOrOwner(olKeyHash, offerId) {
```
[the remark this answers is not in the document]
lnist: The only rationale is what is used in the functions. If only olKeyHash is used, then it only requires that - but for public functions one could argue that olKey versions would be nice. However, we recently added olKeyHash->olKey mapping in Mangrove so we can go both ways. Do you have a preference?

src/strategies/offer_forwarder/abstract/Forwarder.sol
lnist: Fixed.

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
lnist: Exactly! That is what I am currently working on (except I was going to keep the full distribution to allow partial publishing, but we can definitely add a geometric helper that can calculate it on chain.

src/strategies/MangroveOffer.sol
```diff
@@ -278,7 +267,9 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
   {
     // now trying to repost residual
     uint new_gives = __residualGives__(order);
```
lnist, line 269: Introduce possibility to set logPrice via a __residualValues__ function which gives new_gives and logPrice.

src/strategies/MangroveOrder.sol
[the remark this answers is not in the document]
jkrivine: Yes if only the hash is used then give the hash in argument (we need to compute it anyway). If not, then better pass the struct in argument (otherwise that'll force us to use the inverse hash mapping)

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -2,29 +2,25 @@
...
   struct TakerOrder {
-    IERC20 outbound_tkn;
-    IERC20 inbound_tkn;
+    OLKey olKey;
     bool fillOrKill;
-    uint takerWants;
-    uint takerGives;
+    int logPrice;
+    uint fillVolume;
     bool fillWants;
```
jkrivine, line 22: I wonder whether we should not revisit the terminology `fillWants` knowing that we have now a price and a volume.
I would use `int logPrice`, `uint fillVolume`, `bool isBuyOrder`

src/strategies/MangroveOrder.sol
```diff
@@ -37,69 +38,60 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+  ///@param olKey the offer list key.
+  ///@param logPrice the price
+  ///@param gives new amount of `olKey.outbound` offer owner gives
   ///@param offerId the id of the offer to be updated
-  function updateOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint wants, uint gives, uint pivotId, uint offerId)
+  function updateOffer(OLKey memory olKey, int logPrice, uint gives, uint offerId)
     external
     payable
-    onlyOwner(outbound_tkn, inbound_tkn, offerId)
+    onlyOwner(olKey.hash(), offerId)
```
jkrivine, line 62: maybe it would make sense that the hash of the olKey (at least the hash of (outbound, inbound, tickScale) is given as a field of olKey? So we don't need to compute it after the creation of the struct, i.e you get it as `olKey.hash` instead of `olKey.hash()`

src/strategies/MangroveOrder.sol
```diff
@@ -111,11 +103,11 @@ contract MangroveOrder is Forwarder, IOrderLogic {
   function checkCompleteness(TakerOrder calldata tko, TakerOrderResult memory res) internal pure returns (bool) {
     // The order can be incomplete if the price becomes too high or the end of the book is reached.
     if (tko.fillWants) {
```
jkrivine, line 105: `if (tko.isBuyOrder`) { `

src/strategies/MangroveOrder.sol
```diff
@@ -246,45 +239,40 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
-    IERC20 inbound_tkn,
-    TakerOrderResult memory res,
-    uint fund
-  ) internal returns (uint refund) {
-    uint residualWants;
+  function postRestingOrder(TakerOrder calldata tko, OLKey memory olKey, TakerOrderResult memory res, uint fund)
+    internal
+    returns (uint refund)
+  {
+    int offerLogPrice = -tko.logPrice;
```
jkrivine, line 256: This deserves a comment I think. I gather that logPrice is in `[outbound]/[inbound]` dimension. So if I buy at logPrice `p` I should bid at logPrice `-p` ? I see potential misunderstandings here because usually absolute price is in `quote/base`. I think this is a relative price (or some better naming scheme)

jkrivine: Spec is if you pass a sell order at price p, you will post an ask at price p for the remaining volume to sell.
If you pass a buy order at price p, you will post a bid at price p for the remaining volume to buy. In this case the fee is disregarded so that total volume  bought after the resting order is taken would be equal to received volume had the initial market order been completely filled.

src/strategies/MangroveOrder.sol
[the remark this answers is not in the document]
lnist: We never do an inverse lookup in the strat code, so we only get only the hash when we only need the hash.

src/strategies/interfaces/IOrderLogic.sol
lnist: fillWants is the same terminology used in mangrove-core, so we should keep it consistent. But I agree that fillWants is not an intuitive name, but I don't think we've been able to come up with a better name. isBuyOrder is better, but it seem we have tried to keep "buy" and "sell" out of mangrove-core. But maybe this is a better name. Do you think we should use it in core?

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -44,13 +40,12 @@ interface IOrderLogic {
...
-  ///@param inbound_tkn The inbound token of the order.
+  ///@param olKeyHash the hash of the offer list key.
   ///@param taker The address of the taker
   ///@param fillOrKill The fillOrKill that take was called with
-  ///@param takerWants How much the taker wanted
-  ///@param takerGives How much the taker would give
-  ///@param fillWants If true, the market order stopped when `takerWants` units of `outbound_tkn` had been obtained; otherwise, the market order stopped when `takerGives` units of `inbound_tkn` had been sold.
+  ///@param logPrice The price
+  ///@param fillVolume the volume to fill
+  ///@param fillWants if true (buying), the market order stops when `fillVolume` units of `olKey.outbound` have been obtained (fee included); otherwise (selling), the market order stops when `fillVolume` units of `olKey.inbound` have been sold.
```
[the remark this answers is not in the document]
lnist: Yes

src/strategies/MangroveOrder.sol
lnist: @adhusson experimented with memory vs computing and settled on computing. This should be decided in mangrove-core.

src/strategies/interfaces/IOrderLogic.sol
lnist: @adhusson  @espendk : Do you have any opinions or ideas. This should be a change in mangrove-core not mangrove-strat, it should just inherit it.
Note fillIn / fillOut has the caveat that in/out is from makers perspective and fillWants is a taker parameter.

src/strategies/MangroveOrder.sol
lnist: Will do a search and replace if we make the change in core.

src/strategies/MangroveOrder.sol
lnist: Fixed in https://github.com/mangrovedao/mangrove-strats/pull/33/commits/54c36048f0ff519b0e17354a0bd4934c6cf126b2

script/strategies/kandel/KandelPopulate.s.sol
```diff
@@ -12,23 +12,25 @@ import {GeometricKandel} from "mgv_strat_src/strategies/offer_maker/market_makin
...
+import {OLKey} from "mgv_src/MgvLib.sol";
 
 /**
  * @notice Populates Kandel's distribution on Mangrove
  */
 
 /**
  * KANDEL=Kandel_WETH_USDC FROM=0 TO=100 FIRST_ASK_INDEX=50 PRICE_POINTS=100\
- *    RATIO=101 SPREAD=1 INIT_QUOTE=$(cast ff 6 100) VOLUME=$(cast ff 18 0.1)\
+ *    LOG_PRICE_OFFSET=769 SPREAD=1 INIT_QUOTE=$(cast ff 6 100) VOLUME=$(cast ff 18 0.1)\
```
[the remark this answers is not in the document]
lnist: Fixed in https://github.com/mangrovedao/mangrove-strats/pull/33/commits/54c36048f0ff519b0e17354a0bd4934c6cf126b2

src/strategies/interfaces/IOrderLogic.sol
espendk: Conceptually, it's an enum that indicates whether `fillVolume` is `gives` or `wants`. So perhaps instead of `bool` we could have an enum, eg

```Solidity
enum VolumeToFill { Gives, Wants }
```

espendk: and yes, this should IMO be consistent in core and strats

lnist: Any downsides to enums? Because I like it.

jkrivine: When the enum is an argument of an external function (as here) this might create hassle for the SDK? (maybe ethers + abi handles that smoothly).

lnist: I'll close this conversation as resolved - if we do something in core we'll uptake here.

src/strategies/offer_forwarder/OfferForwarder.sol
```diff
@@ -17,80 +18,66 @@ contract OfferForwarder is ILiquidityProvider, Forwarder {
   }
 
   /// @inheritdoc ILiquidityProvider
-  function newOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint wants, uint gives, uint pivotId, uint gasreq)
+  function newOffer(OLKey memory olKey, Tick tick, uint gives, uint gasreq)
```
espendk, line 21: Should we have `*ByVolume` variants of these methods like we have in core?

src/strategies/offer_maker/AaveMaker.sol
```diff
@@ -60,4 +64,21 @@ contract AaveMaker is ITesterContract, OfferMaker, AaveV3Borrower {
...
+    return newOffer(olKey, tick, gives, gasreq);
+  }
+
+  function updateOfferFromVolume(OLKey memory olKey, uint wants, uint gives, uint offerId, uint gasreq)
+    external
+    payable
+  {
+    Tick tick = TickLib.tickFromVolumes(wants, gives);
+    updateOffer(olKey, tick, gives, offerId, gasreq);
+  }
```
espendk, line 83: Why are these not in the base contract `OfferMaker` instead?

src/strategies/offer_maker/DirectTester.sol
```diff
@@ -37,4 +41,21 @@ contract DirectTester is ITesterContract, OfferMaker {
...
+    return newOffer(olKey, tick, gives, gasreq);
+  }
+
+  function updateOfferFromVolume(OLKey memory olKey, uint wants, uint gives, uint offerId, uint gasreq)
+    external
+    payable
+  {
+    Tick tick = TickLib.tickFromVolumes(wants, gives);
+    updateOffer(olKey, tick, gives, offerId, gasreq);
+  }
```
espendk, line 60: Ditto

test/strategies/MgvOrder.t.sol
```diff
@@ -449,72 +497,104 @@ contract MangroveOrder_Test is StratTest {
...
+    Offer offer = mgv.offers(olKey, res.offerId);
+    OfferDetail detail = mgv.offerDetails(olKey, res.offerId);
+    assertEq(offer.gives(), makerGives(sellOrder) / 2, "Incorrect offer gives");
+
+    assertApproxEqRel(offer.wants(), makerWants(sellOrder) / 2, 1e4, "Incorrect offer wants");
     assertEq(offer.prev(), 0, "Offer should be best of the book");
     assertEq(detail.maker(), address(mgo), "Incorrect maker");
   }
 
+  function test_partial_fill_sell_with_resting_order_is_below_density() public {
```
espendk, line 538: Name reads a bit weird. Maybe this is better?
```suggestion
  function test_partial_fill_sell_with_resting_order_below_density() public {
```

espendk (commented):
Wow, epic work!

Added some smaller comments.

I only reviewed the Kandel changes superficially, I trust @jkrivine to do a better job at that.

src/strategies/offer_forwarder/OfferForwarder.sol
lnist: In my opinion not - but OfferForwarder is not part of the audit - so do you mean in ILiquidityProvider?

src/strategies/MangroveOffer.sol
```diff
@@ -247,20 +245,14 @@ abstract contract MangroveOffer is AccessControlled, IOfferLogic {
     data = bytes32(0);
   }
 
-  ///@notice Given the current taker order that (partially) consumes an offer, this hook is used to declare how much `order.inbound_tkn` the offer wants after it is reposted.
+  ///@notice Given the current taker order that (partially) consumes an offer, this hook is used to declare how much `order.olKey.outbound` the offer gives after it is reposted, while also allowing adjustment to the price.
```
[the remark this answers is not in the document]
lnist: We use tick here, so I replaced it by tick.

src/strategies/offer_forwarder/OfferForwarder.sol
espendk: Just feels weird to have `*ByTick` and `*ByVolume` variants in core, but not in strat lib.

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -89,40 +88,38 @@ contract MumbaiMangroveFullTestnetDeployer is Deployer {
...
       reader: reader,
-      tkn1: weth,
-      tkn2: dai,
+      //FIXME: what tick spacing?
+      market: Market({tkn0: weth, tkn1: dai, tickSpacing: 1}),
       tkn1_in_gwei: toGweiOfMatic(prices[2]),
       tkn2_in_gwei: toGweiOfMatic(prices[0]),
       fee: 0
     });
+    //FIXME: what tick spacing?
```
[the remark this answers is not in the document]
lnist: done.

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -79,8 +78,8 @@ contract MumbaiMangroveFullTestnetDeployer is Deployer {
       mgv: mgv,
       gaspriceOverride: 140, // this overrides Mangrove's gasprice for the computation of market's density
       reader: reader,
-      tkn1: dai,
-      tkn2: usdc,
+      //FIXME: what tick spacing?
```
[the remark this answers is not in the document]
lnist: agree

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -65,11 +64,11 @@ contract MumbaiMangroveFullTestnetDeployer is Deployer {
     (KandelSeeder seeder, AaveKandelSeeder aaveSeeder) = new MumbaiKandelSeederDeployer().runWithChainSpecificParams();
 
     // Activate markets
-    IERC20 dai = IERC20(fork.get("DAI"));
-    IERC20 usdc = IERC20(fork.get("USDC"));
-    IERC20 weth = IERC20(fork.get("WETH"));
+    address dai = fork.get("DAI");
+    address usdc = fork.get("USDC");
+    address weth = fork.get("WETH");
```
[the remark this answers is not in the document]
lnist: fixed in https://github.com/mangrovedao/mangrove-strats/pull/33/commits/f2c0a442ab2a48f4e1dea0ebebd37dd6a0b5c336

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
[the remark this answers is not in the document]
lnist: agree

script/deployers/MumbaiMangroveFullTestnetDeployer.s.sol
```diff
@@ -89,40 +88,38 @@ contract MumbaiMangroveFullTestnetDeployer is Deployer {
       mgv: mgv,
       gaspriceOverride: 140,
       reader: reader,
-      tkn1: weth,
-      tkn2: dai,
+      //FIXME: what tick spacing?
```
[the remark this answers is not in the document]
lnist: fixed in f93a00e75bebe40fdad75ad994ccdedbd1db47ca

script/strategies/kandel/deployers/KandelSeederDeployer.s.sol
```diff
@@ -57,36 +58,39 @@ contract KandelSeederDeployer is Deployer {
     fork.set("AavePooledRouter", address(aaveSeeder.AAVE_ROUTER()));
 
     console.log("Deploying Kandel instances for code verification...");
-    IERC20 weth = IERC20(fork.get("WETH"));
-    IERC20 dai = IERC20(fork.get("DAI"));
+    address weth = fork.get("WETH");
+    address dai = fork.get("DAI");
```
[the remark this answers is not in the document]
lnist: fixed in f2c0a442ab2a48f4e1dea0ebebd37dd6a0b5c336

src/strategies/offer_maker/DirectTester.sol
lnist: Fixed in 3569229b591a18aff5b52fdd6fcb18873df1e5d4

src/strategies/offer_maker/AaveMaker.sol
lnist: Fixed in 3569229b591a18aff5b52fdd6fcb18873df1e5d4

espendk: Btw, isn't the naming scheme *ByVolume in core?

lnist: Fixed in 500b04920b6c18599cf1ad9f57a335c1ae719fa2

src/strategies/offer_forwarder/OfferForwarder.sol
lnist: One could argue the audited strat lib should expose all the variants (see, e.g. Direct._updateOffer (and similar for Forwarder)), but that would bloat the already quite big byte-code size.
Due to variants of these function often arising in strats it can quickly explode in complexity. Should Kandel for instance also offer gives/wants based population?
I think we should keep it small, and I am tempted to argue that core should drop the direct support for these variants and instead offer the usage of libs. Core is more static and does not affect downstream strats, so its easier to keep it there, though.

espendk: Alright, I trust your judgment :-)

src/toy_strategies/offer_maker/Amplifier.sol
```diff
@@ -3,12 +3,15 @@ pragma solidity ^0.8.10;
...
-import {MgvLib, MgvStructs} from "mgv_src/MgvLib.sol";
+import {MgvLib, Offer, OfferDetail} from "mgv_src/MgvLib.sol";
+import {TickLib, Tick} from "mgv_lib/TickLib.sol";
 
 contract Amplifier is Direct {
   IERC20 public immutable BASE;
   IERC20 public immutable STABLE1;
   IERC20 public immutable STABLE2;
+  uint public immutable TICK_SCALE1;
+  uint public immutable TICK_SCALE2;
```
espendk, line 14:
```suggestion
  uint public immutable TICK_SPACING1;
  uint public immutable TICK_SPACING2;
```

espendk:
Looks good! Only added some small comments and just realized it’s in toy strategies, so feel free to ignore.

espendk (approved):
🎉

PR #41, opened 2023-09-12
feat(Kandel): Use gives and logPrice for distribution

lnist:


lnist:
Replaced by https://github.com/mangrovedao/mangrove-strats/pull/43

PR #42, opened 2023-09-12
feat: update events to match core

peterMangrove:
This PR will try and update all events in Strats to match the new Mangrove core.

It tries to follow the same principles as in this [PR](https://github.com/mangrovedao/mangrove-core/pull/460)

### DIFF
--- a/src/strategies/MangroveOrder.sol
+++ b/src/strategies/MangroveOrder.sol
@@ -211,11 +211,11 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     return res;
   }
 
-  ///@notice logs `OrderStartSummary`
+  ///@notice logs `MangroveOrderStart`
   ///@param tko the arguments in memory of the taker order
   ///@dev this function avoids loading too many variables on the stack
   function logOrderData(TakerOrder memory tko) internal {
-    emit OrderStartSummary({
+    emit MangroveOrderStart({
       olKeyHash: tko.olKey.hash(),
       taker: msg.sender,
       fillOrKill: tko.fillOrKill,
--- a/src/strategies/MangroveOrder.sol
+++ b/src/strategies/MangroveOrder.sol
@@ -37,12 +37,6 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     }
   }
 
-  ///@notice The expiry of the offer has been set
-  ///@param olKeyHash the hash of the offer list key.
-  ///@param offerId the Mangrove offer id.
-  ///@param date in seconds since unix epoch
-  event SetExpiry(bytes32 indexed olKeyHash, uint offerId, uint date);
-
   ///@inheritdoc IOrderLogic
   ///@dev We also allow Mangrove to call this so that it can part of an offer logic.
   function setExpiry(bytes32 olKeyHash, uint offerId, uint date) public mgvOrOwner(olKeyHash, offerId) {
@@ -135,6 +129,7 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     // POST:
     // * (NAT_USER-`msg.value`, OUT_USER, IN_USER-`takerGives`)
     // * (NAT_THIS+`msg.value`, OUT_THIS, IN_THIS+`takerGives`)
+    logOrderData(tko);
 
     (res.takerGot, res.takerGave, res.bounty, res.fee) = MGV.marketOrderByLogPrice({
       olKey: tko.olKey,
@@ -212,30 +207,22 @@ contract MangroveOrder is Forwarder, IOrderLogic {
     // POST (else)
     // * (NAT_USER+`res.bounty`, OUT_USER+`res.takerGot`, IN_USER-`res.takerGave`)
     // * (NAT_THIS, OUT_THIS, IN_THIS)
-    logOrderData(tko, res);
+
     return res;
   }
 
-  ///@notice logs `OrderSummary`
+  ///@notice logs `OrderStartSummary`
   ///@param tko the arguments in memory of the taker order
-  ///@param res the result of the taker order.
   ///@dev this function avoids loading too many variables on the stack
-  function logOrderData(TakerOrder memory tko, TakerOrderResult memory res) internal {
-    emit OrderSummary({
-      mangrove: MGV,
+  function logOrderData(TakerOrder memory tko) internal {
+    emit OrderStartSummary({
       olKeyHash: tko.olKey.hash(),
       taker: msg.sender,
       fillOrKill: tko.fillOrKill,
       logPrice: tko.logPrice,
       fillVolume: tko.fillVolume,
       fillWants: tko.fillWants,
-      restingOrder: tko.restingOrder,
-      expiryDate: tko.expiryDate,
-      takerGot: res.takerGot,
-      takerGave: res.takerGave,
-      bounty: res.bounty,
-      fee: res.fee,
-      restingOrderId: res.offerId
+      restingOrder: tko.restingOrder
     });
   }
 
--- a/src/strategies/interfaces/IOfferLogic.sol
+++ b/src/strategies/interfaces/IOfferLogic.sol
@@ -12,13 +12,13 @@ interface IOfferLogic is IMaker {
   ///@notice Log incident (during post trade execution)
   ///@param olKeyHash the hash of the offer list key. This is indexed so that RPC calls can filter on it.
   ///@param offerId the Mangrove offer id. This is indexed so that RPC calls can filter on it.
-  ///@param makerData from the maker. Should we index this?
-  ///@param mgvData from Mangrove. Should we index this?
+  ///@param makerData from the maker.
+  ///@param mgvData from Mangrove.
   ///@notice By emitting this data, an indexer can keep track of what indcidents that has happended.
   event LogIncident(bytes32 indexed olKeyHash, uint indexed offerId, bytes32 makerData, bytes32 mgvData);
 
   ///@notice Logging change of router address
-  ///@param router the new router address. Should we index this?
+  ///@param router the new router address.
   ///@notice By emitting this an indexer can keep track of what router is used.
   event SetRouter(AbstractRouter router);
 
--- a/src/strategies/interfaces/IOfferLogic.sol
+++ b/src/strategies/interfaces/IOfferLogic.sol
@@ -10,17 +10,16 @@ import {AbstractRouter} from "mgv_strat_src/strategies/routers/AbstractRouter.so
 
 interface IOfferLogic is IMaker {
   ///@notice Log incident (during post trade execution)
-  ///@param mangrove The mangrove deployment.
-  ///@param olKeyHash the hash of the offer list key.
-  ///@param offerId the Mangrove offer id.
-  ///@param makerData from the maker.
-  ///@param mgvData from Mangrove.
-  event LogIncident(
-    IMangrove mangrove, bytes32 indexed olKeyHash, uint indexed offerId, bytes32 makerData, bytes32 mgvData
-  );
+  ///@param olKeyHash the hash of the offer list key. This is indexed so that RPC calls can filter on it.
+  ///@param offerId the Mangrove offer id. This is indexed so that RPC calls can filter on it.
+  ///@param makerData from the maker. Should we index this?
+  ///@param mgvData from Mangrove. Should we index this?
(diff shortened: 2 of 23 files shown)


src/strategies/interfaces/IOfferLogic.sol
```diff
@@ -10,17 +10,16 @@ import {AbstractRouter} from "mgv_strat_src/strategies/routers/AbstractRouter.so
...
+  ///@param olKeyHash the hash of the offer list key. This is indexed so that RPC calls can filter on it.
+  ///@param offerId the Mangrove offer id. This is indexed so that RPC calls can filter on it.
+  ///@param makerData from the maker. Should we index this?
+  ///@param mgvData from Mangrove. Should we index this?
+  ///@notice By emitting this data, an indexer can keep track of what indcidents that has happended.
+  event LogIncident(bytes32 indexed olKeyHash, uint indexed offerId, bytes32 makerData, bytes32 mgvData);
 
   ///@notice Logging change of router address
-  ///@param router the new router address
+  ///@param router the new router address. Should we index this?
```
lnist, line 21: delete "should we index this".
I don't see a need - this is a per contract setting. If could get all events for all IOfferLogic contracts and _then_ filter , then it would make sense. But we can't.

src/strategies/MangroveOrder.sol
```diff
@@ -212,30 +207,22 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
       fillVolume: tko.fillVolume,
       fillWants: tko.fillWants,
-      restingOrder: tko.restingOrder,
-      expiryDate: tko.expiryDate,
-      takerGot: res.takerGot,
-      takerGave: res.takerGave,
-      bounty: res.bounty,
-      fee: res.fee,
-      restingOrderId: res.offerId
+      restingOrder: tko.restingOrder
```
lnist, line 225: I assume the rest is in OrderComplete for market order - and that is also emitted if no offers are taken.
Note that Mangrove.js may only be listening to OrderStartSummary, so it may need changes.

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -39,37 +39,31 @@ interface IOrderLogic {
...
-  ///@param expiryDate The expiry date take was called with
-  ///@param takerGot How much the taker got
-  ///@param takerGave How much the taker gave
-  ///@param bounty How much bounty was given
-  ///@param fee How much fee was paid for the order
-  ///@param restingOrderId If a restingOrder was posted, then this holds the offerId for the restingOrder
-  event OrderSummary(
-    IMangrove mangrove,
+  ///@notice By emitting this data, an indexer will be able to tell that we are in the context of an mangroveOrder and keep track of what parameters was use to start the order.
+  event OrderStartSummary(
```
espendk, line 50: why not just `OrderStart`? This doesn't feel like a summary...

src/strategies/interfaces/IOfferLogic.sol
```diff
@@ -10,17 +10,16 @@ import {AbstractRouter} from "mgv_strat_src/strategies/routers/AbstractRouter.so
...
-  ///@param olKeyHash the hash of the offer list key.
-  ///@param offerId the Mangrove offer id.
-  ///@param makerData from the maker.
-  ///@param mgvData from Mangrove.
-  event LogIncident(
-    IMangrove mangrove, bytes32 indexed olKeyHash, uint indexed offerId, bytes32 makerData, bytes32 mgvData
-  );
+  ///@param olKeyHash the hash of the offer list key. This is indexed so that RPC calls can filter on it.
+  ///@param offerId the Mangrove offer id. This is indexed so that RPC calls can filter on it.
+  ///@param makerData from the maker. Should we index this?
```
lnist, line 15: Note: I find it unusual to leave comments like "Should we index this?" without a FIXME or TODO  - easy to miss :)

Hopefully there are so few of these that indexing data does not matter, so I would save the gas.

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -39,37 +39,31 @@ interface IOrderLogic {
...
     uint fillVolume,
     bool fillWants,
-    bool restingOrder,
-    uint expiryDate,
-    uint takerGot,
-    uint takerGave,
-    uint bounty,
-    uint fee,
-    uint restingOrderId
+    bool restingOrder
```
espendk, line 57: I take it the removed fields can be inferred from other events?

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -39,37 +39,31 @@ interface IOrderLogic {
...
-  ///@param mangrove The Mangrove contract on which the offer was posted
-  ///@param olKeyHash the hash of the offer list key.
-  ///@param taker The address of the taker
+  ///@param olKeyHash the hash of the offer list key. This could be found by the OrderStart event, but is needed for RPC call. This is indexed so that RPC calls can filter on it.
+  ///@param taker The address of the taker. This could be found by the OrderStart event, but is needed for RPC call. This is indexed so that RPC calls can filter on it.
   ///@param fillOrKill The fillOrKill that take was called with
-  ///@param logPrice The price
-  ///@param fillVolume the volume to fill
+  ///@param logPrice The logPrice of the order. This is not needed for an indexer, as it can get it from the OrderStart event. It is only emitted for RPC calls.
+  ///@param fillVolume the volume to fill. This is not needed for an indexer, as it can get it from the OrderStart event. It is only emitted for RPC calls.
```
lnist, line 46: If RPC calls need these two, then what about the rest of the fields?

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -39,37 +39,31 @@ interface IOrderLogic {
...
-    uint restingOrderId
+    bool restingOrder
   );
 
+  ///@notice The expiry of the offer has been set
+  ///@param olKeyHash the hash of the offer list key. It is indexed so RPC call can filter on it.
+  ///@param offerId the Mangrove offer id. Should this be indexed?
+  ///@param date in seconds since unix epoch
+  ///@notice By emitting this data, an indexer will be able to keep track of the expiry date of an offer.
+  event SetExpiry(bytes32 indexed olKeyHash, uint offerId, uint date);
```
lnist, line 65: index offerId?

src/strategies/interfaces/IOrderLogic.sol
lnist, line 50: Consider indexed restingOrderId

src/strategies/interfaces/IOrderLogic.sol
lnist: Ah, it is the same as newOwnedOffer? hm do we need both?

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
```diff
@@ -15,13 +15,17 @@ abstract contract DirectWithBidsAndAsksDistribution is Direct, HasIndexedBidsAnd
...
   ///@notice logs the end of a call to populate
+  ///@notice By emitting this, an indexer will know that the previous PopulateStart event is over.
   event PopulateEnd();
 
   ///@notice logs the start of a call to retractOffers
+  ///@notice By emitting this, an indexer will be able to know that the following events are in the context of retract.
   event RetractStart();
   ///@notice logs the end of a call to retractOffers
+  ///@notice By emitting this, an indexer will know that the previous RetractStart event is over.
   event RetractEnd();
```
espendk, line 29: These don't need any indexed data for RPC calls to filter on?

src/strategies/routers/integrations/AavePooledRouter.sol
```diff
@@ -25,13 +25,15 @@ contract AavePooledRouter is HasAaveBalanceMemoizer, AbstractRouter {
 
   ///@notice The `aaveManager` has been set.
   ///@param manager the new manager.
+  ///@notice By emitting this data, an indexer will be able to keep track of what manager is used.
   event SetAaveManager(address manager);
```
espendk, line 29: Indexed data for for RPC filtering?

espendk (commented):
Great :-) I added some questions.

Someone with more knowledge of Kandel should review this, it's not clear to me what is needed there.

src/strategies/MangroveOrder.sol
peterMangrove: Agreed, i've already run into that when using the SDK. But i think it is doable

src/strategies/interfaces/IOrderLogic.sol
peterMangrove: Hmm, you might be on to something. If we listen to both `OrderStart` and `OrderStartSummary` (name might change) in SDK, then `logPrice`, `fillVolume` and `fillWants`, are not needed here, as they are already emitted from `OrderStart`. `fillOrKill` and `restingOrder`, are mangroveOrder specific, so we still need them

src/strategies/interfaces/IOrderLogic.sol
peterMangrove: Hmm, good question. Are you thinking that by listening to both `NewOwnedOffer` and `SetExpiry`, then we do not need `offerId` on `SetExpiry`, as we know that we will emit `SetExpiry` after `NewOwnedOffer`, thereby implicitly knowing the OfferId?

src/strategies/interfaces/IOrderLogic.sol
peterMangrove: Hmm, that field has been removed, as it can be found via `NewOwnedOffer`, and it is indexed on that event

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
lnist: I don't think so - it is for a specific kandel instance already. And we would need to use the event numbers to use the information for anything, so I think we are good here. At least I cannot come up with good data to include :)

src/strategies/interfaces/IOrderLogic.sol
peterMangrove: Yes, if we listing to `OrderStartSummary` (will change the name), `OrderStart`,  `SetExpiry`, the `OfferSucces/Fail`'s and `OrderComplete` events, then we will be able to get the same information. It will be a bit more tricky in the SDK, but it will save gas

src/strategies/interfaces/IOfferLogic.sol
peterMangrove: Agreed, it was my own notes and forgot to clean up before asking for review.

src/strategies/interfaces/IOrderLogic.sol
lnist: Sorry I am in doubt what I could have meant here. SetExpiry is only for expiring offers, so it does not replace NewOwnedOffer, and if we reuse an offerId then we need the setExpiry event. So both are needed, and I think offerId should be indexed on both.

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
peterMangrove: Agreed, but could be interesting to try and catch the information doing a populate or retract, by only using RPC calls, to see if it is doable.

src/strategies/interfaces/IOrderLogic.sol
peterMangrove: Agreed. The old name was `OrderSummary` and i moved it up, to the beginning. Maybe `MangroveOrderStart` is better? Mangrove already has an event called `OrderStart`, so it might be confusing to have the same name

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
espendk: Good idea. Let's merge this first version and then adapt if needed

espendk (approved):
Good for me.

I think @lnist  should also review, it looks like there's still some open discussions from his last review.

lnist (approved):

PR #44, opened 2023-09-13
Feat/kandel init new offer

lnist:
3 commits, can be reviewed independently.
Last one is proof of concept.
We still have a givesDist, and we only allow populating _all_ at once. That does not support chunking.
The reason for having a givesDist is that we have worked with both: constant base, constant quote, constant outgoing. Either we need to implement a function for each of these or we take a dist.
To do chunking we can make a from;to version instead of going over all price points.


src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
```diff
@@ -98,6 +98,46 @@ abstract contract GeometricKandel is CoreKandel {
...
 
+  ///@notice sets the log price offset
+  ///@param _logPriceOffset the log price offset.
+  function setLogPriceOffset(int _logPriceOffset) public onlyAdmin {
+    require(int24(_logPriceOffset) == _logPriceOffset, "Kandel/logPriceOffsetTooHigh");
+    logPriceOffset = _logPriceOffset;
+    emit SetLogPriceOffset(_logPriceOffset);
+  }
+
+  ///@notice publishes bids/asks for the distribution given by
```
jkrivine, line 109: incomplete comment

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
```diff
@@ -98,6 +98,46 @@ abstract contract GeometricKandel is CoreKandel {
...
+      }
+      distribution.indices[i] = i;
+      distribution.logPriceDist[i] = logPrice;
+      if (i >= firstAskIndex) {
+        logPrice += _logPriceOffset;
+      } else {
+        logPrice -= _logPriceOffset;
+      }
+    }
+    this.populate{value: msg.value}(distribution, firstAskIndex, parameters, baseAmount, quoteAmount);
```
jkrivine, line 138: can't we make populate use distribution in memory rather than in calldata to avoid this? This is quite gas costly (external call rather than a jump). Especially that on L2 you want to call populate with the explicit distribution only internally

jkrivine (changes requested):

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
lnist: Added more.

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
```diff
@@ -40,49 +40,79 @@ abstract contract DirectWithBidsAndAsksDistribution is Direct, HasIndexedBidsAnd
     uint[] indices;
     int[] logPriceDist;
     uint[] givesDist;
+    bool createDual;
```
lnist, line 43: make parameter

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
```diff
@@ -46,47 +42,56 @@ abstract contract DirectWithBidsAndAsksDistribution is Direct, HasIndexedBidsAnd
...
-  ///@param distribution the distribution of prices for gives of base and quote for indices.
-  ///@param firstAskIndex the (inclusive) index after which offer should be an ask.
+  ///@notice Publishes bids/asks for the distribution in the `indices`. Care must be taken to publish offers in meaningful chunks. For instance, for Kandel an offer and its dual should be published in the same chunk (one being optionally initially dead).
+  ///@param bidDistribution the distribution of prices for gives of quote for indices.
+  ///@param askDistribution the distribution of prices for gives of base for indices.
   ///@param gasreq the amount of gas units that are required to execute the trade.
   ///@param gasprice the gasprice used to compute offer's provision.
-  function populateChunk(Distribution calldata distribution, uint firstAskIndex, uint gasreq, uint gasprice) internal {
+  ///@dev Gives of 0 means create/update and then retract offer (but update price, gasreq, gasprice of the offer)
+  function populateChunk(
```
jkrivine, line 51: Can we call this function `_populateChunk` or `populateChunkInternal` ? I don't like an internal function to seemingly override a public one.

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
```diff
@@ -42,196 +25,240 @@ abstract contract GeometricKandel is CoreKandel {
...
+  ///@dev for Bids, outbound=quote, inbound=base so relative taker price of a a bid is the inverse of the absolute price.
+  ///@dev for Asks, outbound=base, inbound=quote so relative taker price of an ask coincides with absolute price.
+  ///@dev Index0 will contain the ask with the lowest relative price and the bid with the highest relative price. Absolute price is geometrically increasing over indexes.
+  ///@dev logPriceOffset moves an offer relative price s.t. `AskLogPrice_{i+1} = AskLogPrice_i + logPriceOffset` and `BidLogPrice_{i+1} = BidLogPrice_i - logPriceOffset`
+  ///@dev A hole is left in the middle at the size of spread - either an offer or its dual is posted, not both.
+  ///@dev The caller should make sure the minimum and maximum log price does not exceed the MIN_LOG_PRICE and MAX_LOG_PRICE from respectively; otherwise, populate will fail for those offers.
+  ///@dev If type(uint).max is used for `bidGives` or `askGives` then very high or low prices can yield gives=0 (which results in both offer an dual being dead) or gives>=type(uin96).max which is not supported by Mangrove.
+  function createDistribution(
```
jkrivine, line 94: same remark about public/internal overriding an external function. This makes the code hard to read.

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
```diff
@@ -42,196 +25,240 @@ abstract contract GeometricKandel is CoreKandel {
...
-
-    // args.fund = 0; the offers are already provisioned
-    // posthook should not fail if unable to post offers, we capture the error as incidents
-    args.noRevert = true;
-    args.gasprice = memoryParams.gasprice;
-    args.gasreq = memoryParams.gasreq;
+  ///@param bidGives The initial amount of quote to give for all bids. If 0, only book the offer, if type(uint).max then askGives is used as base for bids, and the quote the bid gives is set to according to the price.
+  ///@param askGives The initial amount of base to give for all asks. If 0, only book the offer, if type(uint).max then bidGives is used as quote for asks, and the base the ask gives is set to according to the price.
+  ///@dev This is typically used after a call to `populateFromOffset` to populate the rest of the offers with the same parameters. See that function for further details.
+  function populateChunkFromOffset(
```
jkrivine, line 242: Nice!

jkrivine: This is the function we want to optimize for L2. So it should eventually accept packed calldata.
Low hanging fruit is that `from`, `to`, `firstAskIndex`, `bidGives`, `bidAsk` can definitely hold on a single `uint`

src/strategies/offer_maker/market_making/kandel/abstract/CoreKandel.sol
```diff
@@ -13,39 +13,140 @@ import {TransferLib} from "mgv_lib/TransferLib.sol";
...
   ///@param gasreq the gasreq to use for offers
+  ///@param gasprice the gasprice to use for offers
   ///@param reserveId identifier of this contract's reserve when using a router.
-  constructor(IMangrove mgv, OLKey memory olKeyBaseQuote, uint gasreq, address reserveId)
+  constructor(IMangrove mgv, OLKey memory olKeyBaseQuote, uint gasreq, uint gasprice, address reserveId)
     TradesBaseQuotePair(olKeyBaseQuote)
     DirectWithBidsAndAsksDistribution(mgv, gasreq, reserveId)
-  {}
+  {
+    setGasprice(gasprice);
```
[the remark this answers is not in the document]
lnist: Hm, I think we have always done that in the strats so we kept the pattern. You are right that the initial populate will anyway contain pricepoints and spread, and thus it could also contain gasprice (right now I think the SDK sets it to what it was, but SDK could set it to a default value if 0)

I am ok with removing it, wdyt?

src/strategies/offer_maker/market_making/kandel/abstract/CoreKandel.sol
```diff
@@ -13,39 +13,140 @@ import {TransferLib} from "mgv_lib/TransferLib.sol";
...
+  ///@notice Publishes bids/asks for the distribution in the `indices`. Care must be taken to publish offers in meaningful chunks. For Kandel an offer and its dual should be published in the same chunk (one being optionally initially dead).
+  ///@notice This function is used publicly after `populate` to reinitialize some indices or if multiple transactions are needed to split initialization due to gas cost.
+  ///@notice This function is not payable, use `populate` to fund along with populate.
+  ///@param bidDistribution the distribution of prices for gives of quote for indices.
+  ///@param askDistribution the distribution of prices for gives of base for indices.
+  function populateChunk(Distribution calldata bidDistribution, Distribution calldata askDistribution)
+    external
+    onlyAdmin
+  {
+    populateChunk(bidDistribution, askDistribution, params.gasreq, params.gasprice);
```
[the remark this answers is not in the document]
lnist: I've changed it to have a local parameters.
It will call with 0 gasreq and gasprice if not set. That is up to the user to initialize correctly. First with populate (to also set spread, pricepoints) then populateChunk

src/strategies/offer_maker/market_making/kandel/abstract/CoreKandel.sol
```diff
@@ -13,39 +13,140 @@ import {TransferLib} from "mgv_lib/TransferLib.sol";
...
+    }
+    setParams(parameters);
+
+    depositFunds(baseAmount, quoteAmount);
+
+    populateChunk(bidDistribution, askDistribution, params.gasreq, params.gasprice);
+  }
+
+  ///@notice Publishes bids/asks for the distribution in the `indices`. Care must be taken to publish offers in meaningful chunks. For Kandel an offer and its dual should be published in the same chunk (one being optionally initially dead).
+  ///@notice This function is used publicly after `populate` to reinitialize some indices or if multiple transactions are needed to split initialization due to gas cost.
```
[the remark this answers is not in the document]
lnist: changed to externally.

src/strategies/offer_maker/market_making/kandel/abstract/DirectWithBidsAndAsksDistribution.sol
lnist: Renamed to populateChunkInternal

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
lnist: Alright, I'll do this in a separate commit, but yes I agree. But there are many things that could be packed I think

src/strategies/offer_maker/market_making/kandel/abstract/GeometricKandel.sol
lnist: Hm this is intentionally public so it can be called from SDK to make a calculation. So I don't know what to change. One createDistribution reads storage but saves on parameters, the other doesn't read storage but then requires all parameters.
The first is to calculate when parameters are already set correctly, the other is for when not set or changing parameters.

PR #46, opened 2023-09-13
feat(MangroveOrder): Allow re-use of dead, owned offers to save gas and provision

lnist:
TODO: add tests of reuse and verify guards.

### DIFF
--- a/src/strategies/MangroveOrder.sol
+++ b/src/strategies/MangroveOrder.sol
@@ -264,18 +264,25 @@ contract MangroveOrder is Forwarder, IOrderLogic {
       // partialFill => tko.fillVolume > res.takerGave
       residualGives = tko.fillVolume - res.takerGave;
     }
-    (res.offerId,) = _newOffer(
-      OfferArgs({
-        olKey: olKey,
-        logPrice: offerLogPrice,
-        gives: residualGives,
-        gasreq: offerGasreq(), // using default gasreq of the strat
-        gasprice: 0, // ignored
-        fund: fund,
-        noRevert: true // returns 0 when MGV reverts
-      }),
-      msg.sender
-    );
+    OfferArgs memory args = OfferArgs({
+      olKey: olKey,
+      logPrice: offerLogPrice,
+      gives: residualGives,
+      gasreq: offerGasreq(), // using default gasreq of the strat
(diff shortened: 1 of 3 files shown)


src/strategies/MangroveOrder.sol
```diff
@@ -264,18 +264,25 @@ contract MangroveOrder is Forwarder, IOrderLogic {
...
+    });
+    if (tko.offerId == 0) {
+      (res.offerId,) = _newOffer(args, msg.sender);
+    } else {
+      uint offerId = tko.offerId;
+      require(ownerData[olKey.hash()][offerId].owner == msg.sender, "AccessControlled/Invalid");
+      require(!MGV.offers(olKey, offerId).isLive(), "mgvOrder/offerAlreadyActive");
+      if (_updateOffer(args, offerId) == REPOST_SUCCESS) {
+        res.offerId = offerId;
+      }
```
jkrivine, line 284: to be consistent with the rest of the code we should add:
```solidity
else { 
  res.offerId = 0 ;
}

File: src/strategies/interfaces/IOrderLogic.sol [hunk 2]
```diff
@@ -22,6 +22,7 @@ interface IOrderLogic {
     bool fillWants;
     bool restingOrder;
     uint expiryDate;
+    uint offerId;
```
jkrivine, line 25: missing natspec for `offerId`

jkrivine (changes requested):
A few comments.
Also on the sdk, how hard is it to reuse old offerIds?

src/strategies/interfaces/IOrderLogic.sol
lnist: Fixed

src/strategies/MangroveOrder.sol
lnist: Added - though it is redundant initialization.

lnist:
> A few comments. Also on the sdk, how hard is it to reuse old offerIds?

Via RPC we can filter on this event `event NewOwnedOffer(bytes32 indexed olKeyHash, uint indexed offerId, address indexed owner);` on olKeyHash+owner to get offers for that owner, and then check each if they are live.
Indexer can hand out all offers for owner.

PR #49, opened 2023-09-14
Feat/variable router gasreq

jkrivine:


src/strategies/routers/abstract/MonoRouter.sol
```diff
@@ -0,0 +1,26 @@
...
+
+  ///@inheritdoc AbstractRouter
+  function __routerGasreq__(address) internal view override returns (uint) {
+    return routerGasreq();
+  }
+
+  ///@notice Mono router specific constant gasreq view
+  function routerGasreq() public view returns (uint) {
+    return ROUTER_GASREQ;
+  }
```
maxencerb, line 25: Est-ce qu'on ne peut pas juste définir __routerGasReq__() ? Comme la fonction routerGasReq() retourne le résultat de __routerGasReq() dans `AbstractRouter.sol`

```suggestion
  ///@inheritdoc AbstractRouter
  function __routerGasreq__(address) internal view override returns (uint) {
    return ROUTER_GASREQ;
  }
```

src/strategies/routers/abstract/MultiRouter.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier:	BSD-2-Clause
+pragma solidity ^0.8.10;
+
+import {IERC20} from "mgv_src/MgvLib.sol";
+import {TransferLib} from "mgv_src/strategies/utils/TransferLib.sol";
+import {MonoRouter, AbstractRouter} from "./MonoRouter.sol";
+
+///@title `MultiRouter` instances have reserveId dependant sourcing strategies.
+abstract contract MultiRouter is AbstractRouter {
+  mapping(address => MonoRouter) public routes;
```
lnist, line 10: Should we consider mapping(IERC20 token => mapping address reserveId => MonoRouter) ?
This may enable Aave router for base, but SimpleRouter for Quote in Kandel?

 (note: you can name parameters in mappings now:
`mapping(address reserveId => MonoRouter route) public routes;`

src/strategies/routers/abstract/MonoRouter.sol
jkrivine: All this is syntactic sugar but the idea is:
* `__routerGasreq__` is the hook to override to define router gas usage.
* `routerGasreq()` is a Mono router *public* function that avoids having to call the `routerGasreq(address)` as imposed by `AbstractRouter` 
So your change is fine but you would have to always call `monoRouter.routerGasreq(address(0))` to know the gasreq of the router which I find inelegant.

maxencerb: indeed, didn't see the signature of the function changed from `AbstractRouter.sol` tbh.

maxencerb (approved):

src/strategies/routers/abstract/MultiRouter.sol
jkrivine: OK but in order to be fully generic maybe the mapping should be 
`mapping(bytes32 signature => MonoRouter)` where signature can be `keccak(reserveId, token)` ?

lnist: You can use address0 for no token. Its only a bit more hashing so should be ok.

lnist (approved):

maxencerb (approved):

maxencerb (approved):

lnist (approved):

PR #53, opened 2023-09-19
Feat/dispatcher router

maxencerb:


src/strategies/offer_forwarder/DispatcherForwarder.sol
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier:	BSD-2-Clause
+pragma solidity ^0.8.10;
+
+import {Forwarder, IMangrove, IERC20} from "mgv_strat_src/strategies/offer_forwarder/abstract/Forwarder.sol";
+import {ILiquidityProvider} from "mgv_strat_src/strategies/interfaces/ILiquidityProvider.sol";
+import {AbstractRouter} from "mgv_strat_src/strategies/routers/SimpleRouter.sol";
+import {Dispatcher} from "mgv_strat_src/strategies/routers/integrations/Dispatcher.sol";
+import {MgvLib} from "mgv_src/MgvLib.sol";
+
+contract OfferForwarder is ILiquidityProvider, Forwarder {
```
jkrivine, line 10: We try also to name the contract X when the file containing it is X.sol (that helps). Maybe you could name the file `OfferDispatcher.sol` and the contract `OfferDispatcher`

src/strategies/offer_forwarder/DispatcherForwarder.sol
```diff
@@ -0,0 +1,125 @@
...
+    AbstractRouter router_ = router();
+    router_.bind(address(this));
+    if (deployer != msg.sender) {
+      setAdmin(deployer);
+      router_.setAdmin(deployer);
+    }
+  }
+
+  /// @inheritdoc ILiquidityProvider
+  function newOffer(IERC20 outbound_tkn, IERC20 inbound_tkn, uint wants, uint gives, uint pivotId, uint gasreq)
```
jkrivine, line 21: Maybe we don't need that version since offers are routing dependant, so you always want to call `offerGasreq()` to know?

src/strategies/routers/integrations/AaveMemoizer.sol
```diff
@@ -40,13 +49,26 @@ contract AaveMemoizer is AaveV3Borrower {
...
 
   ///@notice contract's constructor
   ///@param addressesProvider address of AAVE's address provider
   ///@param interestRateMode  interest rate mode for borrowing assets. 0 for none, 1 for stable, 2 for variable
   constructor(address addressesProvider, uint interestRateMode) AaveV3Borrower(addressesProvider, interestRateMode) {}
 
+  ///@notice fetches and optionally memoizes the owner of `this` contract in additional data
+  ///@param m the memoizer
+  ///@return owner the owner of the funds
+  function getOwnerAddress(Memoizer memory m) internal view returns (address owner) {
```
jkrivine, line 63: maybe just `function owner(Memoizer memory m)` to match the other memo getters.
An alternative to adding `owner` to the memo is to require an additional argument to functions that require it. Would be more elegant and avoid too much implicits.
So `balanceOf(m)` would be `balanceOf(m, owner)` etc. This would be consistent with the fact that you pushed `maxGettableUnderlying` in the Memoizer code (and this function requires an extra `target` arg)

src/strategies/routers/integrations/Dispatcher.sol
```diff
@@ -0,0 +1,128 @@
...
+  /// @param reserveId The reserveId to call the function on
+  /// @param token The token to call the function on
+  /// @param data The data to call the function with
+  function callRouterSpecificFunction(bytes4 selector, address reserveId, IERC20 token, bytes calldata data)
+    external
+    onlyBound
+  {
+    address router = routerSpecificFunctions[selector];
+    require(router != address(0), "Dispatcher/SelectorNotSet");
+    (bool success,) = router.delegatecall(abi.encodeWithSelector(selector, reserveId, token, data));
```
jkrivine, line 112: what happens if you delegatecall address(0)? Will this call return `success=false` or just throw?

src/strategies/routers/integrations/dispatched/AaveDispatchedRouter.sol
```diff
@@ -0,0 +1,136 @@
...
+
+/// @title `AaveDispatchedRouter` is a router contract for Aave used by the `Dispatcher` contract.
+/// @dev No tokens should be directly sent to this contract
+/// @dev This contract is to be used by the `Dispatcher` contract.
+contract AaveDispatchedRouter is MonoRouter, AaveMemoizer {
+  /// @notice Holds the storage key for this contract specific storage
+  bytes32 internal immutable STORAGE_KEY;
+
+  /// @notice The maximum credit line to be redeemed
+  /// * 100 means we can redeem or borrow 100% of the credit line
```
jkrivine, line 17: Maybe a bit more specific (I guess that number only matters if the user has debt)

src/strategies/routers/integrations/dispatched/AaveDispatchedRouter.sol
```diff
@@ -0,0 +1,136 @@
...
+  /// * 0 means we can't redeem or borrow anything
+  uint internal immutable MAX_CREDIT_LINE = 100;
+
+  /// @notice Data for a reserve <=> token pair
+  /// @param deposit_on_push Whether to deposit on push
```
jkrivine, line 22: Discussion to be had: I think we can implement the "deposit on push" as a particular route for the token to be pushed. For instance an offer can have aave route for token A et wallet route for token B. If offer gives A and receives B it will pull from AAVE and deposit to the wallet.
This requires a change in the way offerGasreq() is computed though

src/strategies/offer_forwarder/DispatcherForwarder.sol
maxencerb: I don't know but if not, we will not conform to the `ILiquidityProvider` ?

src/strategies/routers/integrations/Dispatcher.sol
maxencerb: It sets success to true and won't throw from my tests, not sure why

src/strategies/routers/integrations/dispatched/AaveDispatchedRouter.sol
maxencerb: Agree on that point, 2 routers seems a great option, just the deployment cost to take into account.

src/strategies/offer_forwarder/DispatcherForwarder.sol
jkrivine: Right.

src/strategies/routers/integrations/dispatched/AaveDispatchedRouter.sol
jkrivine: But the routers will be deployed anyway. Suppose you have `SimpleRouter` and `AaveRouter`. You can then chose which one you want to use for pull and push

maxencerb: That's a great idea, just have to change the multirouter data structure that points to a single `MonoRouter` for all operations.

PR #58, opened 2023-09-22
feat(Kandel): Reduce bytecode size

lnist:
1. Look at each commit individually
2. "[Dont check PooledRouter every call](https://github.com/mangrovedao/mangrove-strats/pull/58/commits/ca989bc74cbd15716cfbd303dfd8c78fc2fe0cf6)" may be controversial, but it saves gas and bytes and I would argue its up to the caller. And code still fails in depost/withdraw, etc., just with different error. But it can be reverted and we are still below limit.
3. Introduction of library has to be tested with foundry, if it wont work easily like this, then we'll create an extra contract to call, but the idea will be the same.


jkrivine (approved):
lgtm

src/strategies/offer_maker/market_making/kandel/AaveKandel.sol
```diff
@@ -39,14 +39,6 @@ contract AaveKandel is GeometricKandel {
     require(!isOverlying, "AaveKandel/cannotTradeAToken");
   }
 
-  ///@notice returns the router as an Aave router
```
jkrivine, line 42: Why don't you just remove the require in this case?

jkrivine (approved):

src/strategies/offer_maker/market_making/kandel/AaveKandel.sol
lnist: Good point. done

PR #72, opened 2023-10-01
fix(core): Upgrade core to new version with mwei gasprice, rounding up wants, and 127 bits for gives

lnist:


src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -89,17 +89,17 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
     returns (uint gasprice, uint leftover)
   {
     unchecked {
-      uint num = (offerGasbase + gasreq) * 10 ** 9;
+      uint num = (offerGasbase + gasreq) * 1e6;
       // pre-check to avoid underflow since 0 is interpreted as "use Mangrove's gasprice"
       require(provision >= num, "mgv/insufficientProvision");
       // Gasprice is eventually a uint16, so too much provision would yield a gasprice overflow
```
adhusson, line 95: a uint26

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -89,17 +89,17 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
       // Reverting here with a clearer reason
-      require(provision < type(uint16).max * num, "Forwarder/provisionTooHigh");
+      require(provision < (1 << 26) * num, "Forwarder/provisionTooHigh");
```
adhusson, line 97: Before it was computing `((1<<16)-1) * num` and now it's computing `(1 << 26) * num`. I don't really understand what is being checked though.

test/strategies/unit/OfferLogic.t.sol
```diff
@@ -254,7 +254,7 @@ contract OfferLogicTest is StratTest {
       gasreq: gasreq
     });
 
-    mgv.setGasprice(type(uint16).max);
+    mgv.setGasprice(1 << 26 - 1);
```
adhusson, line 257: This sets gasprice to `1<<25`, not `(1<<26)-1`

adhusson (commented):
I haven't checked that all the 1e9 have been properly replaced with either 1e6 or 1e12, do the tests fail if they are switched?

src/strategies/offer_forwarder/abstract/Forwarder.sol
lnist: good catch! Found a few more like these.

src/strategies/offer_forwarder/abstract/Forwarder.sol
lnist: Thanks, you are right, I've fixed it. The check here is to do the following:
revert with  Forwarder/provisionTooHigh
instead of  mgv/writeOffer/gasprice/tooBig
Reason being that Forwarder uses all the provision offered by a user of the forwarder to provision the offer. Thus a gasprice is derived, and gasprice is not a direct concept for the user, so this is a better error message.

test/strategies/unit/OfferLogic.t.sol
lnist: This is a test :D so it just sets the gasprice high and verifies that it did not provide enough provision. Any high value would do. But I've added parenthesis to get the expected value.

lnist: Fixed in a58a8656e14890761f86d5e9dc016bff31054ef5

src/strategies/offer_forwarder/abstract/Forwarder.sol
lnist: Fixed in a58a8656e14890761f86d5e9dc016bff31054ef5

src/strategies/offer_forwarder/abstract/Forwarder.sol
lnist: Fixed in a58a8656e14890761f86d5e9dc016bff31054ef5

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -147,7 +147,7 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
     (uint gasprice, uint leftover) = deriveAndCheckGasprice(args);
 
     // the call below cannot revert for lack of provision (by design)
-    // it may still revert if `args.fund` yields a gasprice that is too high (mangrove's gasprice is uint16)
+    // it may still revert if `args.fund` yields a gasprice that is too high (mangrove's gasprice is uint26)
```
jkrivine, line 150: `gasprice` must hold on 26 bits.

src/strategies/offer_maker/market_making/kandel/abstract/CoreKandel.sol
```diff
@@ -45,9 +46,9 @@ abstract contract CoreKandel is DirectWithBidsAndAsksDistribution, TradesBaseQuo
   ///@param stepSize in amount of price points to jump for posting dual offer.
   ///@param pricePoints the number of price points for the Kandel instance.
   struct Params {
-    uint16 gasprice;
+    uint32 gasprice;
```
jkrivine, line 49: // must hold on 26 bits

jkrivine (approved):
Maybe one little thing to update in the comments but lgtm

PR #82, opened 2023-10-09
Feat/permit2 makers

jkrivine:
This PR proposes to allow strats made with the strat lib (either Direct or Forwarder) to use permit2 approval.
Permit2 based approvals for Direct is not implemented and left to the designer's choice (in order not to impose additional gas cost if the strat does not want to use permit2).
* 112890b23cbfd16f0f0f651be6914f1f74f08cba introduces the main changes
* the following commit propagates changes to tests and examples

### DIFF
--- a/src/toy_strategies/interfaces/ITesterContract.sol
+++ b/src/toy_strategies/interfaces/ITesterContract.sol
@@ -9,10 +9,12 @@ import {IERC20} from "mgv_lib/IERC20.sol";
 interface ITesterContract is ILiquidityProvider {
   function tokenBalance(IERC20 token, address reserveId) external view returns (uint);
 
-  function newOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq)
+  function newOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint gasreq, bool usePermit2)
     external
     payable
     returns (uint offerId);
 
-  function updateOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint offerId, uint gasreq) external payable;
+  function updateOfferByVolume(OLKey memory olKey, uint wants, uint gives, uint offerId, uint gasreq, bool usePermit2)
+    external
+    payable;
 }
--- a/src/toy_strategies/offer_forwarder/AmplifierForwarder.sol
+++ b/src/toy_strategies/offer_forwarder/AmplifierForwarder.sol
@@ -92,7 +92,8 @@ contract AmplifierForwarder is Forwarder {
         gasreq: offerGasreq(), // SimpleRouter is a MonoRouter
         gasprice: 0, // ignored
         fund: args.fund1,
-        noRevert: false
+        noRevert: false,
+        usePermit2: false
       }),
       msg.sender
     );
@@ -110,7 +111,8 @@ contract AmplifierForwarder is Forwarder {
         gasreq: offerGasreq(),
         gasprice: 0, // ignored
         fund: args.fund2,
-        noRevert: false
+        noRevert: false,
+        usePermit2: false
       }),
       msg.sender
     );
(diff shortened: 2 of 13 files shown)


src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -266,13 +275,16 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
     // telling router one is requiring `amount` of `outTkn` for `owner`.
     // because `pull` is strict, `pulled <= amount` (cannot be greater)
     // we do not check local balance here because multi user contracts do not keep more balance than what has been pulled
 
     ApprovalInfo memory approvalInfo;
-    uint pulled = router().pull(IERC20(order.olKey.outbound_tkn), owner, amount, true, approvalInfo);
+    if (data.usePermit2) {
+      approvalInfo.approvalType = ApprovalType.Permit2Approval;
+    }
+    uint pulled = router().pull(IERC20(order.olKey.outbound_tkn), data.owner, amount, true, approvalInfo);
```
maxencerb, line 287: There is something wrong with the current logic. We are only. a partial `ApprovalInfo` object if we use permit2.

Currently, if we use permit2 with an offer forwarder on pull, this is what the code does :

1. pushes successfully
2. uses the `ApprovalTransferLib.sol` lib to pull
3. Use the `Permit2TransferLib.sol` lib function `transferTokenFromWithPermit2` that will try to permit our forwarder to pull from the reserve, but our signature is `0x0` as we didn't pass anything
4. From now it failed but should transfer tokens to forwarder

The only thing that should differ is `permit2.transferFrom(...)` instead of `token.transferFrom(...)` if we use permit2.

The approval should have been given in advance by the reserve right ?

maxencerb (changes requested):

src/strategies/offer_forwarder/abstract/Forwarder.sol
Louis-Amas: Did you test it ?
What I remember implementing was that if `approvalInfo.spender `is `0x0` we directly call `permit2.transferFrom` which expect the use approve using permit2 in advance.
So here when Jean is onlyu setting up approvalType it means that spender and signature will be `0x0` meaning that we wil directly call permit2.transferFrom which will succeed if 

- The user approve ERC20 -> Permit2
- The user approved the router using Permit2 by using an EOA to `approve` or a contract calling `permit` with the user signature

>The approval should have been given in advance by the reserve right ?
yes

src/strategies/interfaces/ILiquidityProvider.sol
```diff
@@ -14,15 +14,20 @@ interface ILiquidityProvider is IOfferLogic {
   ///@param gives the amount of inbound tokens the offer maker gives for a complete fill
   ///@param gasreq the gas required by the offer logic
   ///@return offerId the Mangrove offer id.
-  function newOffer(OLKey memory olKey, Tick tick, uint gives, uint gasreq) external payable returns (uint offerId);
+  function newOffer(OLKey memory olKey, Tick tick, uint gives, uint gasreq, bool usePermit2)
```
Louis-Amas, line 17: This was my concern, every strat needs to be aware of the existence of Permit2. It's probably OK, but for me, it's feel wrong.

src/strategies/offer_forwarder/abstract/Forwarder.sol
maxencerb: Just assumed, should have looked into permit2 before then. The function sig  is very mysterious then that's why, I'll try and run it

src/strategies/interfaces/ILiquidityProvider.sol
Louis-Amas: Discussed here I'm convinved now :)
https://mangroveitops.slack.com/archives/C05V909PKH7/p1696947297903319

src/strategies/offer_forwarder/abstract/Forwarder.sol
jkrivine: Yes the code above assumes the signature has been already given to permit2

Louis-Amas (approved):

PR #84, opened 2023-10-10
feat(gasreq): Measure gasreq for MangroveOrder and Kandel

lnist:
The measurements are currently only for the generic foundry test-environment with our test tokens.

Measurements should be made for deployments when we have the checklists for those giving us optimization values and specific tokens to measure.

Note: Currently gasreq for router is ignored for Kandel and not separately measured - it is difficult to attribute the gasusage to the router. In practice it has not impact as the router is not changed.


jkrivine (approved):
great !

PR #86, opened 2023-10-11
Changing gasreq

maxencerb:


src/strategies/routers/abstract/MultiRouter.sol
```diff
@@ -16,8 +16,23 @@ abstract contract MultiRouter is AbstractRouter {
...
+  }
+
+  ///@inheritdoc AbstractRouter
+  function __routerPushGasreq__(IERC20 inbound_tkn, address reserveId) internal view override returns (uint) {
+    return routes[inbound_tkn][reserveId].ROUTER_PUSH_GASREQ();
+  }
+
+  ///@inheritdoc AbstractRouter
+  function __routerPullGasreq__(IERC20 outbound_tkn, IERC20, address reserveId) internal view override returns (uint) {
+    return routes[outbound_tkn][reserveId].ROUTER_PULL_GASREQ();
```
lnist, line 35: I'm a bit worried about all these calls just to get the gasreq - a gasreq that may anyway be difficult to separate from the gasreq of the strat. So maybe all of this logic is actually strat-logic. The strat can then know if it has multiple routers, what their interactions could be, a gasreq matrix, etc. And a simple strat could just return a constant.

However, in order to have some progress I am also fine with this change in general, we can then reconsider a bit down the line.

One issue though: You get a constant here, shouldn't you be calling routerPullgasreq of the router?

lnist: Also, these kind of changes would be better to have targeting the feat/advancedRouting branch so we know what is reviewed and what is not when we get to the stargate branch. And yes, it is a bit confusing with all these branches - as soon as develop is audited we can target that one again.

PR #97, opened 2023-10-18
Feat/remove default gasreq

jkrivine:
In order to prepare future development of strats where gas cost for executing an offer depends on routing, we remove here default gasreq from strats.

Main changes of this PR are in commit 2c84273e8b7ebbbe which does the following:
* removes `offerGasreq()` from `MangroveOffer` class (and related functions)
* removes `routerGasreq()` from `AbstractRouter` class  (and related functions)
* Kandel strats have a default gasreq for offers stored in `params` which accounts for the totality of the gas requirement to execute an offer.
* MangroveOrder no longer has a default gasreq, and the gas requirement is now an arguement of `TakerOrder` struct. This is to anticipate next development of MangroveOffer which will have offer specific custom sourcing strategies (and therefore variable gascosts).


src/strategies/routers/abstract/MultiRouter.sol
maxencerb, line 1: This abstraction is going to be deleted anyway for our implementation

maxencerb (approved):

script/strategies/kandel/deployers/PolygonKandelSeederDeployer.s.sol
```diff
@@ -19,9 +19,8 @@ contract PolygonKandelSeederDeployer is Deployer {
     return new KandelSeederDeployer().innerRun({
       mgv: IMangrove(fork.get("Mangrove")),
       addressesProvider: fork.get("Aave"),
-      aaveKandelGasreq: 200_000,
-      kandelGasreq: 200_000,
-      aaveRouterGasreq: 280_000
+      aaveKandelGasreq: 628_000,
```
lnist, line 22: FYI these are not "correct" in the sense they have to be measured to the chain we want to deploy to for the tokens we want to use with the optimization settings we deploy with.

lnist: Also - should we remove these default gasreqs completely and let is be seeder input? MangroveOrder in SDK needs to know a sensible gasreq, so maybe Kandel should do the same? This would remove gasreq completely from deployment.

src/strategies/MangroveOrder.sol
```diff
@@ -253,7 +253,7 @@ contract MangroveOrder is Forwarder, IOrderLogic {
       olKey: olKey,
       tick: residualTick,
       gives: residualGives,
-      gasreq: offerGasreq(), // using default gasreq of the strat
+      gasreq: tko.restingOrderGasreq, // using default gasreq of the strat
```
lnist, line 256: comment is stale

src/strategies/offer_maker/market_making/kandel/Kandel.sol
```diff
@@ -14,7 +14,7 @@ contract Kandel is GeometricKandel {
   ///@param gasreq the gasreq to use for offers
   ///@param reserveId identifier of this contract's reserve when using a router.
   constructor(IMangrove mgv, OLKey memory olKeyBaseQuote, uint gasreq, address reserveId)
```
lnist, line 16: gasreq is still passed in but unused?

src/strategies/routers/abstract/MultiRouter.sol
lnist: If MultiRouter is deleted then I suggest we also delete it from develop so auditors don't spend time on it (they should not as it is not part of the scope, but if we can delete it then it is super clear :) )

src/strategies/offer_maker/market_making/kandel/AaveKandel.sol
```diff
@@ -52,12 +51,13 @@ contract AaveKandel is GeometricKandel {
 
   ///@notice Sets the AaveRouter as router and activates router for base and quote
   ///@param router_ the Aave router to use.
-  function initialize(AavePooledRouter router_) external onlyAdmin {
+  ///@param gasreq the gas required to execute an offer of this Kandel strat
```
lnist, line 54: Is the idea that the gasreq is coupled to the router, so therefore we set it here? That kinda makes sense, but we could also just require a call to setGasreq. I'm ok with this though.

lnist (changes requested):

src/strategies/interfaces/IOrderLogic.sol
```diff
@@ -24,6 +24,7 @@ interface IOrderLogic {
     bool restingOrder;
     uint expiryDate;
     uint offerId;
+    uint restingOrderGasreq;
```
lnist, line 27: Add natspec

lnist, line 27: Allowing the user to specify gasreq was removed during the earlier audit. See https://docs.mangrove.exchange/assets/files/ChainSecurity_Mangrove_Association_ADDMA_MangroveOrder_audit-7e289d0c705233f1d69d419d7689cab5.pdf section 6.1 and 6.2

I think the fix in https://github.com/mangrovedao/mangrove-core/pull/70/commits ("addressing 5.1") is part of the fix, but later we also did mangrovedao/mangrove-core#103 which removed the ability to set gasreq for MangroveOrder. @jkrivine : Why did we do that?

src/strategies/routers/abstract/MonoRouter.sol
```diff
@@ -1,23 +0,0 @@
-// SPDX-License-Identifier:	BSD-2-Clause
-pragma solidity ^0.8.10;
-
-import {IERC20} from "@mgv/lib/IERC20.sol";
-import {AbstractRouter} from "./AbstractRouter.sol";
-
-///@title `MonoRouter` instances have a sourcing strategy which is reserveId and caller independent.
-///@dev `routerGasreq(address reserveId)` is thus a constant function.
-abstract contract MonoRouter is AbstractRouter {
```
lnist, line 9: We need to update AUDIT_TRAIL if this is to be included in the audit now (you are targeting develop?) - both regarding removal of gasreq stuff, and the AUDIT_TRAIL mentions MonoRouter but now it is gone :)

script/strategies/kandel/deployers/PolygonKandelSeederDeployer.s.sol
jkrivine: OK it's more a debate of deployment procedure than code, but agreed.

src/strategies/offer_maker/market_making/kandel/Kandel.sol
jkrivine: no it is, gasreq is set at the end of the constructor.

src/strategies/interfaces/IOrderLogic.sol
jkrivine: Yes but we implemented a fix that allows us to modify gasreq in Forwarder so it's fine to do it here. The issue was that `_updateOffer` in forwarder strats needs to take gasreq into account to potentially free some wei for the owner in case the new gasreq is lower than the old one.

lnist: Agree, so let's leave it at that.

lnist (approved):

maxencerb (approved):

PR #105, opened 2023-10-24
Merge core master changes to strats master

lnist:


espendk (approved):
lgtm

Noting though that it's slightly weird that some addresses are in both mangrove-core and mangrove-strats. But that's not something that's introduced here, so should be fixed here.

PR #116, opened 2023-11-02
ops(deploy): Deploy contracts to Mumbai (v0.0.2-3)

espendk:
In order to get the contracts on-chain asap, this deployment was done using the old deployment process despite it not being adapted to the new multi-repo setup.

This means that addresses and deployment logs for the core contracts are included in the mangrove-strats repo and npm package. In future depoyments, we should avoid this.

Note that hard-coded prices have been used to initialize the markets that use the tokens we have deployed, since no price oracle is available for those.


dontrolle (approved):
👍

PR #135, opened 2023-11-16
Add simple aave logic

maxencerb:


src/strategies/routing_logic/SimpleAaveLogic.sol
```diff
@@ -0,0 +1,59 @@
...
+  ///@notice contract's constructor
+  ///@param addressesProvider address of AAVE's address provider
+  ///@param interestRateMode  interest rate mode for borrowing assets. 0 for none, 1 for stable, 2 for variable
+  constructor(address addressesProvider, uint interestRateMode) AaveMemoizer(addressesProvider, interestRateMode) {}
+
+  ///@inheritdoc AbstractRouter
+  function __pull__(RL.RoutingOrder memory routingOrder, bool strict) internal virtual override returns (uint) {
+    Memoizer memory m;
+
+    uint amount = strict ? routingOrder.amount : balanceOf(routingOrder.token, m, routingOrder.fundOwner);
```
jkrivine, line 19: this should be `overlyingBalanceOf` otherwise call will return if `fundOwner` has 0 underlying but everything in overlying.

jkrivine: Semantics of `strict` is:
transfer at most `amount` if `strict` and all the `token` you can otherwise. If you want to implement the "transfer underlying first" policy then if !strict you should transfer
1. the whole balance of tokens from `fundOwner`
2. the whole aave deposit of tokens (transfer the whole balance of aToken from fundOwner and then redeem)
If strict then:
1. amount_0 = min(amount, token.balanceOf(fundOwner))
2. amount_1 = redeem from aave min(amount-amount_0, 0)
3. transfer amount_0 + amount_1

Yes we need to make sure we want the "transfer underlying first" policy. Did we commit to doing this?

src/strategies/routing_logic/SimpleAaveLogic.sol
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  ///@inheritdoc AbstractRouter
+  function __push__(RL.RoutingOrder memory routingOrder) internal virtual override returns (uint pushed) {
+    Memoizer memory m;
+    // just in time approval of the POOL in order to be able to deposit funds
+    _approveLender(routingOrder.token, routingOrder.amount);
+    uint leftToPush = routingOrder.amount;
+    // tries to repay existing debt
+    if (debtBalanceOf(routingOrder.token, m, routingOrder.fundOwner) > 0) {
```
jkrivine, line 41: `debtBalanceOf` is quite costly we need to make sure we also want the "repay first" policy.

src/strategies/routing_logic/SimpleAaveLogic.sol
```diff
@@ -0,0 +1,59 @@
...
+      uint repaid = _repay(routingOrder.token, leftToPush, routingOrder.fundOwner);
+      leftToPush -= repaid;
+    }
+    // supplies the rest
+    if (leftToPush > 0) {
+      bytes32 reason = _supply(routingOrder.token, leftToPush, routingOrder.fundOwner, true);
+      require(reason == bytes32(0), "AaveLogic/SupplyFailed");
```
jkrivine, line 48: what about `require(reason == bytes32(0), string(abi.encodePacked(reason)))` so as not to loose Aave's message?

src/strategies/routing_logic/SimpleAaveLogic.sol
```diff
@@ -0,0 +1,59 @@
...
+    }
+    return routingOrder.amount;
+  }
+
+  ///@inheritdoc AbstractRouter
+  function balanceOfReserve(RL.RoutingOrder calldata routingOrder) public view virtual override returns (uint balance) {
+    Memoizer memory m;
+    balance = overlyingBalanceOf(routingOrder.token, m, routingOrder.fundOwner);
+    balance += balanceOf(routingOrder.token, m, routingOrder.fundOwner);
```
jkrivine, line 57: to be corrected if we do not implement the "underlying first" policy. Otherwise all good. Maybe just add a comment to say that balanceOfReserve might not be entirely available in case of a liquidity crisis on AAVE

jkrivine (changes requested):

src/strategies/routing_logic/SimpleAaveLogic.sol
maxencerb: No I think we should stick to only overlying, especially in a simple strat like this. It was just an error from my side

src/strategies/routing_logic/SimpleAaveLogic.sol
maxencerb: Indeed, for this simple strat, we should then keep only lending. Repay has a better for Aave APR though, to consider later.

src/strategies/routing_logic/SimpleAaveLogic.sol
maxencerb: In this case the last param is `noRevert` so It will throw inside Aave

src/strategies/routing_logic/SimpleAaveLogic.sol
maxencerb: indeed

src/strategies/routing_logic/SimpleAaveLogic.sol
```diff
@@ -0,0 +1,47 @@
...
+    if (amount == 0) {
+      return 0;
+    }
+
+    require(
+      TransferLib.transferTokenFrom(routingOrder.token, routingOrder.fundOwner, msg.sender, amount),
+      "SimpleAaveLogic/TransferFailed"
+    );
+
+    uint redeemed = _redeem(routingOrder.token, amount, msg.sender);
```
jkrivine, line 30: There is an issue with the `_redeem` function of `AAveMemoizer` which does not catch revert as `_supply` for instance. For uniformity it would be great to have `_redeem` returning `bytes32` with a `noRevert` argument.
The reason is that when you do external calls inside offerLogic you want to have a strict control of reverts (which make your trade fail).

src/strategies/routing_logic/SimpleAaveLogic.sol
```diff
@@ -0,0 +1,47 @@
...
+    return redeemed;
+  }
+
+  ///@inheritdoc AbstractRouter
+  function __push__(RL.RoutingOrder memory routingOrder) internal virtual override returns (uint pushed) {
+    // just in time approval of the POOL in order to be able to deposit funds
+    _approveLender(routingOrder.token, routingOrder.amount);
+    _supply(routingOrder.token, routingOrder.amount, routingOrder.fundOwner, false);
```
jkrivine, line 38: same here I would use the `noRevert` flag to true and let caller decide it wants to revert

jkrivine (changes requested):

jkrivine (approved):
OK modulo call to _redeem in __pull__ in SimpleAaveLogic should have `noRevert=false`

PR #139, opened 2023-11-18
Feat/amplifier offers

jkrivine:


src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+      // Setting logic to push inbound tokens offer
+      (vars.proxy,) = ROUTER_FACTORY.instantiate(msg.sender, ROUTER_IMPLEMENTATION);
+      vars.routingOrder.token = args.inboundTkns[i];
+      vars.routingOrder.olKeyHash = vars.olKeyHash;
+      vars.routingOrder.offerId = vars.offerId;
+      SmartRouter(address(vars.proxy)).setLogic(vars.routingOrder, args.inboundLogics[i]);
+
+      // Setting logic to pull outbount tokens for the same offer
+      vars.routingOrder.token = args.outbound_tkn;
+      SmartRouter(address(vars.proxy)).setLogic(vars.routingOrder, args.outboundLogic);
```
maxencerb, line 136: Reduce gas cost a little bit by doing a check on address before writing to heap, calling and storing 0 in router (reduce cost by ~200 per loop and ~3000 if no sourcing strat at all)

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+
+contract MangroveAmplifier is ExpirableForwarder {
+  ///@notice offer bundle identifier
+  uint private __bundleId;
+
+  ///@notice maps a `bundleId` to the set of inbound_tokens of the bundle
+  mapping(uint bundleId => IERC20[] inbound_tkns) private __inboundTknsOfBundleId;
+
+  ///@notice maps a `bundleId` to the set of pairs (tickSpacing, offerId) of the bundle
+  mapping(uint bundleId => uint[2][] ticks_offerIds) private __ticks_offerIdsOfBundleId;
```
maxencerb, line 32: bundling in a uint[2] is the same as using a struct, and easier to understand.

Furthermore, we should keep bundle => inbound_tkns in the same mapping as the second one because of the storage layout of dynamic arrays.

With this storage layout, [we would need to write the arrays length twice instead of once 
](https://docs.soliditylang.org/en/v0.8.17/internals/layout_in_storage.html#mappings-and-dynamic-arrays
)

Doing something like this would save ~20k gas regardless of the number of token

```solidity
struct BundleSingleOfferInfo {
  uint tickSpacing;
  uint offerId;
  IERC20 inbound_tkn;
}

mapping(uint bundleId => BundleSingleOfferInfo[] bundle_offers_info) private __bundleOffersInfo;
```

This specific order on struct is great to keep the uints having a single storage slot and then we a a partial storage slot with the address.

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+  ///@param factory the router proxy factory used to deploy or retrieve user routers
+  ///@param deployer The address of the admin of `this` at the end of deployment
+  constructor(IMangrove mgv, RouterProxyFactory factory, address deployer)
+    ExpirableForwarder(mgv, factory, new SmartRouter())
+  {
+    _setAdmin(deployer);
+  }
+
+  struct HeapVarsNewBundle {
+    uint[2][] ticks_offerIds;
```
maxencerb, line 62: Having an array of struct is the same here as above, same in gas but easier to understand

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+  /// * `provision` is the portion of `msg.value` that should be allocated to the provision of the i^th offer
+  /// * `tick` is the tick spacing parameter that charaterizes the offer list to which the offer should be posted.
+  ///@param inboundLogics the logics to manage liquidity targetting for each offer of the bundle. Use `AbstractRouter(address(0))` for simple routing.
+  struct BundleArgs {
+    IERC20 outbound_tkn;
+    uint outVolume;
+    AbstractRouter outboundLogic; // AbstractRouter(address(0)) for SimpleRouter behavior.
+    uint expiryDate; // 0 for no expiry
+    IERC20[] inboundTkns;
+    uint[4][] params; // params[0]: inVolume, params[1] = gasreq, params[2] = provision, params[3] = tick
```
maxencerb, line 89: fixed size array and struct have same layout storage in memory so make a struct for ease of use.

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+    AbstractRouter[] inboundLogics;
```
maxencerb, line 90: maybe we could bundle all these arguments in a single struct as they are supposed to be the same size in array. 

Because these are calldata, values set to 0 will cost less + we will only set the length of the dynamic once according to abi specification reducing calldata size by 64 bytes.

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+  mapping(bytes32 olKeyHash => mapping(uint offerId => uint bundleId)) private __bundleIdOfOfferId;
+
+  ///@notice logs beginning of bundle creation
+  ///@param bundleId bundle identifier
+  event InitBundle(uint indexed bundleId);
+
+  ///@notice logs end of bundle creation
+  ///@dev to know which offer are part of the bundle one needs to track `NewOwnedOffer(olKeyHash, offerId, owner)` emitted in between `InitBundle` and `EndBundle`
+  ///@dev the expiry date of the bundle is given by the logs `SetExpiry(olKeyHash, offerId, expiryDate)` of each offer of the bundle (all dates are the same).
+  event EndBundle();
```
maxencerb, line 49: Subject to change if we keep this.

1. keep it

If we keep this, this means that we think other offers could have been created during the same tx. So we would need to specify EndBundle with the bundleId ?

2. drop it

If we add no args, we assume we will only do this during our tx or even create multiple bundles in which case we will know when to stop because we will be emitting InitBundle again.

Dropping it is a ~375 gas cost less.

The only reason to add this would be having a tx creating a bundle and a standalone offer created. Is it a usecase.

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+  ///@param ticks_offerIds the array of elements `[tick_i, offerId_i]` where `tick_i` is the tick spacing of the offer list in which `offerId_i` lives.
+  ///@param inbound_tkns the array of inbound tokens of each member of the bundle to update
+  ///@param outboundVolume the new volume that each offer of the bundle should now offer
+  ///@param updateExpiry whether the update also changes expiry date of the bundle
+  ///@param expiryDate the new date (if `updateExpiry` is true) for the expiry of the offers of the bundle. 0 for no expiry
+  function _updateBundle(
+    IERC20 outbound_tkn,
+    uint offerId,
+    uint[2][] memory ticks_offerIds,
+    IERC20[] memory inbound_tkns,
```
maxencerb, line 186: bundling these two within an array of struc as well and dropping the fixed size array

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+      SmartRouter(address(vars.proxy)).setLogic(vars.routingOrder, args.inboundLogics[i]);
+
+      // Setting logic to pull outbount tokens for the same offer
+      vars.routingOrder.token = args.outbound_tkn;
+      SmartRouter(address(vars.proxy)).setLogic(vars.routingOrder, args.outboundLogic);
+
+      vars.olKeyHash = vars.olKey.hash();
+      if (args.expiryDate != 0) {
+        _setExpiry(vars.olKeyHash, vars.offerId, args.expiryDate);
+      }
```
maxencerb, line 141: It's too bad here we will write expiry date in memory 3 times. Maybe we should have an expiry date for the whole bundle. As such, maybe not using the `ExpirableForwarder`. Indeed, we have a mapping from offer to bundle already, so we don't need to have expiry for each offer.

It would save ~20k gas per offer.

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+          args.olKey = olKey_i;
+          args.tick = offer_i.tick(); // same price
+          args.gives = outboundVolume; // new volume
+          args.gasreq = offerDetail_i.gasreq();
+          args.noRevert = true;
+          // call below will retract the offer without reverting if update fails (for instance if the density it too low)
+          _updateOffer(args, ticks_offerIds[1][i]);
+          if (updateExpiry) {
+            _setExpiry(olKey_i.hash(), ticks_offerIds[1][i], expiryDate);
+          }
```
maxencerb, line 214: if we have an expiry on a per bundle basis, same as above, we would save ~20k gas per offer for updating

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,308 @@
...
+  ///@dev updating offer bundle in makerExecute rather than in makerPosthook to avoid griefing: an adversarial offer logic could collect the bounty of the bundle by taking the not yet updated offers during its execution.
+  ///@dev the risk of updating the bundle during makerExecute is that any revert will make the trade revert, so one needs to make sure it does not happen (by catching potential reverts, assuming no "out of gas" is thrown)
+  function __get__(uint amount, MgvLib.SingleOrder calldata order) internal override returns (uint) {
+    // this will use offer owner's router to pull `amount` to this contract
+    uint missing = super.__get__(amount, order);
+
+    // we know take care of updating the other offers that are part of the same bundle
+    uint bundleId = __bundleIdOfOfferId[order.olKey.hash()][order.offerId];
+    uint[2][] memory ticks_offerIds = __ticks_offerIdsOfBundleId[bundleId];
+    IERC20[] memory inbound_tkns = __inboundTknsOfBundleId[bundleId];
```
maxencerb, line 289: Bundling in struct

src/strategies/MangroveAmplifier.sol
jkrivine: Ha but the reason I am doing this is precisely because you cannot write an array of structs in currently solidity...

src/strategies/MangroveAmplifier.sol
jkrivine: Yes precisely. What happens if a user creates a bundle and then a regular offer?

jkrivine: besides 375 gas is probably not important at that stage

src/strategies/MangroveAmplifier.sol
jkrivine: Yes I was thinking of an expiry per bundle as well and possibly setting the outbound logic for the whole bundle as well. But is it worth the pain knowing that you pay the extra gas upfront only (when you create the bundle) but not duing the offer execution (which is sort of what matters)

src/strategies/MangroveAmplifier.sol
jkrivine: Hum but now it seems you can. So yes of course if it does I will revert to using structs here instead.

maxencerb: This works, just we have to take a different approach as to how to write it to storage, because it does not support converting memory array of struct to storage array of struct. We have to push the members of the struct directly to storage. Here is a working PoC I've made on remix.

```solidity
// SPDX-License-Identifier: GPL-3.0
pragma solidity >=0.8.2 <0.9.0;

contract Test {

    struct TestStruct {
        uint amount;
        uint decimals;
        address token;
    }

    mapping(uint bundleId => TestStruct[] tokens) public tokensMapping;

    function setMappingValue() public {
        TestStruct[] storage tokens = tokensMapping[0];
        tokens.push(TestStruct({
            amount: 1 ether,
            decimals: 1 ether,
            token: address(0)
        }));
        tokens.push(TestStruct({
            amount: 12 ether,
            decimals: 1 ether,
            token: address(0)
        }));
    }

}
```

src/strategies/MangroveAmplifier.sol
maxencerb: 1. logic per bundle

No we should not do that because some logic could be specific to some tokens (e.g. no market on Aave for the selected token).

2. bundle expiry

For the example of having 3 inbounds, so 3 offers, it would save 40k upfront + 40k per modifying offer. I agree it's not optimizing taking the offer but it would not cost more for the taker. I will write a small PR, I don't think there is much modifications to make.

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,318 @@
...
+        vars.routingOrder.olKeyHash = vars.olKeyHash;
+        vars.routingOrder.offerId = vars.bundledOffer.offerId;
+        SmartRouter(address(vars.proxy)).setLogic(vars.routingOrder, vr[i].inboundLogic);
+      }
+
+      // Setting logic to pull outbount tokens for the same offer
+      if (address(fx.outboundLogic) != address(0)) {
+        vars.routingOrder.token = fx.outbound_tkn;
+        SmartRouter(address(vars.proxy)).setLogic(vars.routingOrder, fx.outboundLogic);
+      }
```
maxencerb, line 138: - If we set all logics to 0, are we sure the proxy is going to be created ?
- if we have no inbound logic, the vars.proxy is going to be set to address(0) once setting the outbound logic

Maybe we should put line 127 : 

```solidity
(vars.proxy,) = ROUTER_FACTORY.instantiate(msg.sender, ROUTER_IMPLEMENTATION);
```

before the loop. Plus, it will save some call to the router factory and insure the router proxy is created (or creates it) when creating a bundle

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,318 @@
...
+            // - Mangrove is dead
+            // we then retract the offer (note the offer list cannot be locked at this stage)
+            bytes32 reason = _updateOffer(args, bundle[i].offerId);
+            if (reason != REPOST_SUCCESS) {
+              // we do not deprovision, owner funds can be retrieved on a pull basis later on
+              _retractOffer(olKey_i, bundle[i].offerId, false);
+            }
+            if (updateExpiry) {
+              _setExpiry(olKey_i.hash(), bundle[i].offerId, expiryDate);
+            }
```
maxencerb, line 231: We can put a continue statement after retract offer, setting expiry after if we retract is not necessary. (I don't know if `continue` exists in solidity but at least an `else if`)

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,318 @@
...
+          olKey: vars.olKey,
+          tick: TickLib.tickFromVolumes(vr[i].inVolume, fx.outVolume),
+          gives: fx.outVolume,
+          gasreq: vr[i].gasreq,
+          gasprice: 0, // ignored in Forwarder strats
+          fund: vars.provision,
+          noRevert: false // revert if unable to post
+        }),
+        msg.sender
+      );
```
maxencerb, line 123: Only because I don't know the core protocol v2 very well, isn't there any market lock happening in here ? For example, if there was a `marketOrder` happening before creating new orders and this market order triggers one of the other market we want to enter ? Just like in the updateBundle function ?

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,318 @@
...
+    returns (uint freeWei)
+  {
+    for (uint i; i < bundle.length; i++) {
+      if (bundle[i].offerId != skipId) {
+        OLKey memory olKey_i = OLKey({
+          outbound_tkn: address(outbound_tkn),
+          inbound_tkn: address(bundle[i].inbound_tkn),
+          tickSpacing: bundle[i].tick
+        });
+        freeWei += _retractOffer(olKey_i, bundle[i].offerId, deprovision);
```
maxencerb, line 274: Here we don't check if the market is locked ? I mean if called by the EAO no problem. But if called inside the `__get__` that is a problem no ?

src/strategies/MangroveAmplifier.sol
jkrivine: The strat never creates newOffer during offer logic execution. So this public function is essentially called externally. So if it is called in a context where the offer list is locked, it is the expected behavior that the function should revert.

src/strategies/MangroveAmplifier.sol
jkrivine: Same here. This is used in the context of an external call aiming at retracting the bundle not during the execution of an offerLogic.

maxencerb: yes it is on line 323

src/strategies/MangroveAmplifier.sol
```diff
@@ -0,0 +1,332 @@
...
+    require(bundleExpiryDate == 0 || bundleExpiryDate > block.timestamp, "MgvAmplifier/expiredBundle");
+    return super.__lastLook__(order);
+  }
+
+  ///@notice bundle wide expiry date setter
+  ///@param bundleId the id of the bundle whose expiry date is to be set
+  ///@param date the date of expiry (use 0 for no expiry)
+  ///@dev and offer logic will renege if either the offer's expiry date is passed or it belongs to a bundle whose expiry date has passed.
+  function _setBundleExpiry(uint bundleId, uint date) internal {
+    _setExpiry(0, bundleId, date);
```
maxencerb, line 101: We have to make sure the indexer will be able to catch this. The indexer would have to query the state of the blockchain for every offer. I like the idea of using the same function for expiry.

But, maybe we can trigger all subsequent events meaning triggering the events inside the for loop, and then modifying the deadline bundlewise.

Something is also that if someone tries to set a longer deadline offerwise than bundlewise, it is going to prefer the bundlewise deadline. Is this what we want ?

maxencerb (changes requested):

maxencerb (approved):

src/strategies/MangroveAmplifier.sol
jkrivine: good point not sure about this.

PR #155, opened 2023-12-05
feat(MangroveOrder): add gas overhead test

Louis-Amas:
Actual change in mangrove.js
https://github.com/mangrovedao/mangrove.js/pull/1651


test/strategies/MgvOrder-gasoverhead.t.sol
```diff
@@ -0,0 +1,89 @@
...
+      fillWants: false,
+      fillVolume: volume,
+      tick: Tick.wrap(10),
+      restingOrder: true,
+      expiryDate: block.timestamp + 10000,
+      offerId: 0,
+      restingOrderGasreq: GASREQ // overestimate
+    });
+
+    sellResult = mangroveOrder.take{value: 1 ether}(sellOrder);
```
lnist, line 75: also prank taker here?

lnist, line 75: add gas measurement using _gas and gas_

Consider putting these tests beside the other gastests so the gas-measurement script picks them up, and then also use the printDescription like the other gasreq tests do.

Louis-Amas: Nope, this test have been made so that we do not it to prank taker. The objective was to simplify the reading.

lnist (approved):

PR #162, opened 2023-12-13
Feat/skip market order option review

jkrivine:


src/strategies/TakerOrderLib.sol
```diff
@@ -10,6 +10,9 @@ pragma solidity >=0.8.10;
 /// @param PO Post only -> This order will not execute immediately against the market.
 /// @param IOC Immediate or cancel -> This order must be filled immediately at the limit price or better. If the full order cannot be filled, the unfilled portion is canceled.
 /// @param FOK Fill or kill -> This order must be filled in its entirety immediately at the limit price or better, or it is entirely canceled. There is no partial fulfillment.
+
+///@dev This enum is used to represent the type of a taker order.
+///@dev invariant must be that TakerOrderType < 3 is a resting order, and >= 3 is a market order.
```
maxencerb, line 15:
```suggestion
///@dev invariant must be that TakerOrderType < 3 will post a resting order, otherwise it won't
```

maxencerb: This invariant as is sounds like only one or the other, thus the suggestion

PR #163, opened 2023-12-14
docs(license): Change license to MIT for all code except 3rd party code

espendk:


dontrolle (approved):
Approving - checked that all relevant files have been updated from AGPL-3.0

PR #209, opened 2024-01-18
Response to nethermind audit

maxencerb:


src/strategies/MangroveAmplifier.sol
```diff
@@ -188,26 +188,56 @@ contract MangroveAmplifier is RenegingForwarder {
...
+    // We check the owner of the bundle ID by checking the owner of the first offer of the bundle
+    owner = ownerOf(olKeyHash, offerId);
+
+    // 2 bundles can share the same offer id for a given outbound token.
+    // if outbound token is manipulated, the function could resolve with a different bundle owner.
+    // Example:
+    // * User 1 posts bundle 1 on WBTC/DAI and WBTC/USDC with offerIds 1 and 2 respectively.
+    // * User 2 posts bundle 2 on WETH/DAI and WETH/USDC with offerIds 1 and 2 respectively.
+    // * if we pass bundle id of 1 and outbound token of WETH, the function will return user 2 as owner.
+
```
jkrivine, line 225: Yes but the bundle only contains the inbound token so the remainder of the code is using the "manipulated" `outbound_tkn` resulting in updating the offer of the caller anyway.

jkrivine, line 225: Yes but the bundle only contains the inbound token so the remainder of the code is using the "manipulated" `outbound_tkn` resulting in updating the offer of the caller anyway.
I think the proper attack scenario is slightly more complex here is one:
```
Alice: bundle [1] WBTC/DAI [id 1] and WBTC/USDT [id 1]
Bob: bundle [2] WETH/DAI [id 1] 
Carl: bundle [3] WETH/USDT [id 1]

Bob: updateBundle(bundleId:1, outbound:WETH)
check: ownerOf(id:1, inbound: DAI, outbound: WETH) == bob --> True
update(id:1, outbound: WETH, inbound: DAI) <- the offer of bob.
update(id:1, outbound: WETH, inbound: USDT) <- the offer of Carl!
```
Maybe after the patch it's not worth putting the whole scenario here

src/strategies/MangroveAmplifier.sol
```diff
@@ -188,26 +188,56 @@ contract MangroveAmplifier is RenegingForwarder {
...
+    // To solve this issue, we do a reverse check where we check the resulting bundle id from olKeyHash and offerId.
+    // An **invariant** is that a bundle id can only be associated to one offer id on a given Offer List.
+    // To follow the same example as above:
+    // * if we pass bundle id of 1 and outbound token of WETH, the function will return user 2 as owner.
+    // * the mapping __bundleIdOfOfferId[olKeyHash][offerId] will either return 0 or a a bundle id different than 2.
+    // * this will result in a revert as the bundle id of 1 is not associated to offer id 1 on the WBTC/DAI offer list.
+    require(__bundleIdOfOfferId[olKeyHash][offerId] == bundleId, "MgvAmplifier/invalidBundleId");
```
jkrivine, line 232: to save a storage read (hot) we can also use a `bytes32` as key for the bundle map. You would build the key in the `newBundle` function by hashing the concatenation of the fresh bundle id and the outbound token. Something like:
```
///@notice maps a `bundleHash` to a set of bundled offers
  mapping(bytes32 bundleKey => BundledOffer[]) private __bundles;
  uint private __bundleId;
[..]
function newBundle(FixedBundleParams calldata fx, VariableBundleParams[] calldata vr)
    public
    payable
    returns (bytes32 freshBundleKey)
  {
    HeapVarsNewBundle memory vars;

    freshBundleKey = keccak256(abi.encode(fx.outbound_tkn, __bundleId++));
    
    emit InitBundle(freshBundleKey);
[...]
```

src/strategies/offer_forwarder/abstract/Forwarder.sol
```diff
@@ -253,18 +253,21 @@ abstract contract Forwarder is IForwarder, MangroveOffer {
...
       status = bytes32(bytes(reason));
     }
+    // if `deprovision` is set to false, `freeWei` is 0 because `retractOffer` will return 0
+    if (freeWei > 0) {
+      // pulling free wei from Mangrove to `this`
+      require(MGV.withdraw(freeWei), "Forwarder/withdrawFail");
+      // resetting pending returned provision
+      // calling code must now assign freeWei to owner
+      od.weiBalance = 0;
+    }
```
jkrivine, line 270: Yes. Possible issue with the old code is when an offer logic calls `_retractOffer` on a locked offer list, the call will throw but the freeWei would still be positive. Note that this is not an issue with current strats that have no funds locally (so there would be nothing to withdraw). But this could be an issue with someone writing a strat on top of Forwarder

jkrivine (changes requested):

src/strategies/MangroveAmplifier.sol
maxencerb: The problem is that we would have to know the bundle id passed to the bundle key in order to validate the outbound token, so either add to the return of newBundle and the `InitBundle` event the `bundleId` and `bundleKey`, to pass these arguments as well when updating the offer. That would make it quite unreadable.

src/strategies/MangroveAmplifier.sol
```diff
@@ -188,26 +188,63 @@ contract MangroveAmplifier is RenegingForwarder {
...
+    // * Bob: updateBundle(bundleId:1, outbound:WETH)
+    // * check: ownerOf(id:1, inbound: DAI, outbound: WETH) == bob --> True
+    // * update(id:1, outbound: WETH, inbound: DAI) <- the offer of bob.
+    // * update(id:1, outbound: WETH, inbound: USDT) <- the offer of Carl!
+
+    // To solve this issue, we do a reverse check where we check the resulting bundle id from olKeyHash and offerId.
+    // An **invariant** is that a bundle id can only be associated to one offer id on a given Offer List.
+    // To follow the same example as above:
+    // * if we pass bundle id of 1 and outbound token of WETH, the function will return user 2 as owner.
+    // * the mapping __bundleIdOfOfferId[olKeyHash][offerId] will either return 0 or a a bundle id different than 2.
```
Louis-Amas, line 237: typo there is two "a"

src/strategies/offer_forwarder/RenegingForwarder.sol
```diff
@@ -111,12 +111,8 @@ contract RenegingForwarder is Forwarder {
...
 
     require(cond.date == 0 || block.timestamp < uint(cond.date), "RenegingForwarder/expired");
-    require(cond.volume == 0 || order.takerWants < uint(cond.volume), "RenegingForwarder/overSized");
-    if (cond.volume != 0) {
-      // if a max volume is set and we are not reneging, we set it back to 0 so that we don't check again and save some gas.
-      // note this is OK since reposting the offer in MangroveOffer.__posthookSuccess__ will update the volume according to `__residualValues__`
-      _setReneging(olKeyHash, order.offerId, cond.date, 0);
-    }
+    require(cond.volume == 0 || order.takerWants <= uint(cond.volume), "RenegingForwarder/overSized");
+
```
Louis-Amas, line 115: I didn't fully follow, but why aren't we calling ?
```
_setReneging
```

maxencerb: When we called it here, we removed the reneging volume before computing the residual values to repost an offer. But, this volume is used as a temp value as the max volume, so when computing residual values, we would have used higher values than the `cond.volume`. Thus, we moved the resetting inside the `_updateOffer` to make sure we don't overpromis.

src/strategies/MangroveAmplifier.sol
```diff
@@ -350,11 +359,11 @@ contract MangroveAmplifier is RenegingForwarder {
...
-    BundledOffer[] memory bundle = __bundles[bundleId];
+    BundledOffer[] memory bundle = __bundles[bundleId][IERC20(order.olKey.outbound_tkn)];
     // if funds are missing, the trade will fail and one should retract the bundle
     // we also retract the bundle if there is no more outbound volume to offer (this avoids reverting of updateOffer for a too low density)
     // otherwise we update the bundle to the new volume
-    uint newOutVolume = order.offer.gives() - order.takerWants;
+    (uint newOutVolume,) = __residualValues__(order);
     if (missing == 0 && newOutVolume > 0) {
       _updateBundle(bundle, IERC20(order.olKey.outbound_tkn), olKeyHash, newOutVolume);
     } else {
```
jkrivine, line 369: This does not address issue 7.7: the call to `_retractBundle` below will be undone when `missing > 0`. We should test missing == 0 below and maybe add a retractBundle in `posthookFallback` if we want to retract the whole bundle if liquidity is missing.

jkrivine (dismissed):

Louis-Amas (approved):

Louis-Amas (approved):

PR #241, opened 2024-02-16
feat(blast): support blast native functions

Louis-Amas:


maxencerb (dismissed):

maxencerb (dismissed):

src/strategies/routers/RouterProxy.sol
```diff
@@ -1,13 +1,14 @@
...
 
+import {BlastGasAndYieldClaimable} from "@mgv-strats/src/strategies/utils/BlastGasAndYieldClaimable.sol";
 import {AbstractRouter} from "./abstract/AbstractRouter.sol";
 
 /// @title Minimalistic Mangrove Router Proxy
 /// @notice A proxy contract that delegates calls to an instance of an AbstractRouter contract.
 ///         It does not allow updates of implementation contract.
 /// @dev it assumes the IMPLEMENTATION uses first storage slot to hold the admin address (this is guaranteed as long as `AbstractRouter` is `AccessControlled`)
-contract RouterProxy {
+contract RouterProxy is BlastGasAndYieldClaimable(msg.sender) {
```
maxencerb, line 11: msg.sender is router proxy factory, so maybe we hardcode an address ?

Louis-Amas: Let's make routerProxy ownalbe

maxencerb (approved):

PR #242, opened 2024-02-19
Add blast support to strats

maxencerb:


script/strategies/amplifier/deployers/MangroveAmplifierDeployer.s.sol
```diff
@@ -41,11 +41,7 @@ contract MangroveAmplifierDeployer is Deployer {
     uint64 nonce = vm.getNonce(broadcaster());
     broadcast();
     // See MangroveAmplifierGasreqBaseTest description for calculation of the gasreq.
-    if (forMultisig) {
-      mgvAmp = new MangroveAmplifier{salt: salt}(mgv, routerProxyFactory, routerImplementation);
-    } else {
-      mgvAmp = new MangroveAmplifier(mgv, routerProxyFactory, routerImplementation);
-    }
+    mgvAmp = deployMangroveAmplifier(mgv, routerProxyFactory, routerImplementation);
```
espendk, line 44: I think the `broadcast()` call should be inside the function. Otherwise, it's difficult to reason about.

Same in the MangroveOrder deployer.

src/strategies/vendor/blast/Blast.sol
espendk, line 1: As discussed, it would be great to use the Blast contracts etc. from the mangrove-core repo, instead of duplicating them. But that can till later when there's time.

espendk (commented):
Looks overall good 👍 Added a few comments, one of which I think should be addressed now, the other can wait.

espendk (approved):
LGTM!

PR #292, opened 2025-02-10
test commit

maxencerb:


src/strategies/offer_maker/market_making/kandel/ERC4626Kandel.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    token.transfer(recipient, amount);
+  }
+
+  ///@notice Allows the admin to withdraw native tokens.
+  ///@param amount The amount of native tokens to withdraw.
+  ///@param recipient The recipient of the native tokens.
+  function adminWithdrawNative(uint amount, address recipient) public onlyAdmin {
+    payable(recipient).transfer(amount);
+  }
```
maxencerb: there is an issue here. The goal with morpho especially is that the router is going to hold the vault shares and so get the rewards, so we should be able to withdraw the rewards ffrom the vault, not only from the kandel contract itself.

src/strategies/offer_maker/market_making/kandel/ERC4626KandelSeeder.sol
```diff
@@ -0,0 +1,62 @@
...
+      KANDEL_GASREQ,
+      Direct.RouterParams({
+        routerImplementation: router, // using ERC4626 router to source liquidity
+        fundOwner: owner,
+        strict: liquiditySharing
+      })
+    );
+    // Allowing newly deployed Kandel to bind to the ERC4626Router
+    router.bind(address(kandel));
+    router.setAdmin(msg.sender);
```
maxencerb: The admin should be the kandel in the schema that I sent you.

Basically, the deployer of the kandel is the admin of the kandel, and the kandel is the admin of the deployed router.

So in this schema, in order to withdraw token from router, you have to go through the kandel contract, same goes for the possible rewards token

Here is a pseudo-code: 

kandel.withdrawToken(amount) onlyAdmin
  local_balance = token.balanceOf(this)
  from_local = min(amount, local_balance)
  token.transfer(from_local, receiver)
  router.withdraw(amount-from_local, to=receiver)
  
 And here is the schema as reference:
 
 
image

src/strategies/routers/integrations/ERC4626Router.sol
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  function adminWithdrawTokens(IERC20 base, IERC20 quote, IERC20 token, uint amount, address recipient)
+    external
+    onlyAdmin
+  {
+    require(token != base && token != quote, "Cannot withdraw underlying tokens");
+    require(address(token) != address(vaults[base]), "Cannot withdraw ERC4626 vault address(tokens) for address base");
+    require(address(token) != address(vaults[quote]), "Cannot withdraw ERC4626 vault tokens for quote");
```
maxencerb: fix errror message different

src/strategies/routers/integrations/ERC4626Router.sol
```diff
@@ -0,0 +1,134 @@
...
+      if (shares > 0) {
+        uint maxRedeemable = oldVault.maxRedeem(address(this));
+        require(maxRedeemable >= shares, "ERC4626Router/maxRedeemExceeded");
+        oldVault.redeem(maxRedeemable < shares ? maxRedeemable : shares, address(this), address(this));
+      }
+    }
+
+    // Set the new vault
+    vaults[token] = vault;
+  }
```
maxencerb: here you only redeem, you do not redeposit. You should add this to the function as well as some tests linked to this.

By the way, we can set a vault to address(0). It means that the asset are not deposited into any vaults

src/strategies/routers/integrations/ERC4626Router.sol
```diff
@@ -0,0 +1,134 @@
...
+    // it needs to also flush tokens with a contract-local 0 amount.
+    // token[0/1] can be address(0) if using this function for only one token
+    if (address(token0) != address(0)) {
+      pushed0 = __push__(RL.createOrder({fundOwner: msg.sender, token: token0}), amount0);
+      _deposit(token0);
+    }
+    if (address(token1) != address(0)) {
+      pushed1 = __push__(RL.createOrder({fundOwner: msg.sender, token: token1}), amount1);
+      _deposit(token1);
+    }
```
maxencerb: maybe the deposit can happen inside of the __push__ and __pull__ ? Or is it a choice ?

src/strategies/routers/integrations/ERC4626Router.sol
```diff
@@ -0,0 +1,134 @@
...
+    RL.RoutingOrder memory routingOrder = RL.createOrder({fundOwner: msg.sender, token: token});
+    return __pull__(routingOrder, amount, true);
+  }
+
+  function pushAndDeposit(IERC20 token0, uint amount0, IERC20 token1, uint amount1)
+    external
+    onlyBound
+    returns (uint pushed0, uint pushed1)
+  {
+    // Push will fail for amount of 0, but since this function is only called for the first maker contract in a chain
```
maxencerb: why will it fail with a 0 amount ?

src/strategies/routers/integrations/MorphoVaultRouter.sol
```diff
@@ -0,0 +1,62 @@
...
+  /// @param amount The amount of rewards to claim
+  /// @param proof The proof for claiming rewards
+  /// @param receiver The address that will receive the claimed rewards
+  /// @dev Only callable by the admin
+  function claimRewardsForToken(IERC20 token, uint amount, bytes32[] calldata proof, address receiver)
+    external
+    onlyAdmin
+  {
+    MORPHO_REWARD_DISTRIBUTOR.claim(msg.sender, address(token), amount, proof);
+    token.transfer(receiver, amount);
```
maxencerb: use the transfer lib instead

src/strategies/routers/integrations/ERC4626Router.sol
Solthodox: Its supposed to be a choice

src/strategies/routers/integrations/ERC4626Router.sol
Solthodox: no, will remove the comment

src/strategies/offer_maker/market_making/kandel/ERC4626KandelSeeder.sol
```diff
@@ -0,0 +1,68 @@
...
+
+  ///@notice the ERC4626 router.
+  ERC4626Router public immutable ERC4626_ROUTER;
+
+  ///@notice constructor for `ERC4626KandelSeeder`. Initializes an `ERC4626Router` with this seeder as admin.
+  ///@param mgv The Mangrove deployment.
+  ///@param erc4626KandelGasreq the total gasreq to use for executing a kandel offer
+  constructor(IMangrove mgv, uint erc4626KandelGasreq) AbstractKandelSeeder(mgv, erc4626KandelGasreq) {
+    ERC4626_ROUTER = new ERC4626Router();
+  }
```
maxencerb: deploy one router per kandel inside _deployKandel

src/strategies/routers/integrations/MorphoVaultRouter.sol
```diff
@@ -0,0 +1,62 @@
...
+
+  /// @notice Sets the vault for a specific token
+  /// @param token The token for which to set the vault
+  /// @param vault The vault to set for the token
+  /// @dev Only callable by the admin
+  function setVaultForToken(IERC20 token, IERC4626 vault) external override onlyAdmin {
+    require(MORPHO_FACTORY.isMorphoVault(address(vault)), "MorphoRouter/notMorpho");
+    vaults[token] = vault;
+    emit VaultSet(token, vault);
+  }
```
maxencerb: We should call super.setVaultForToken(...)

maxencerb (commented):
Create a `MorphoKandelSeeder` as well, so jsut the basic erc4626kandel with the morphoRouter

PR #295, opened 2025-03-14
Cantina audit fixes

Solthodox:


src/strategies/routers/integrations/ERC4626Router.sol
```diff
@@ -68,14 +86,16 @@ contract ERC4626Router is AbstractRouter {
...
     (bool s,) = recipient.call{value: amount}("");
     require(s, "ERC4626Router/adminWithdrawNativeFailed");
+    emit AdminNativeWithdrawal(amount, recipient);
   }
 
   /// @notice Sets the vault for a token
   /// @param token The token to set the vault for
   /// @param vault The vault to set
-  function setVaultForToken(IERC20 token, IERC4626 vault) public virtual onlyAdmin {
+  function setVaultForToken(IERC20 token, IERC4626 vault, uint minAssetsOut) public virtual onlyAdmin {
```
maxencerb: Maybe we could also check slippage for the amount of newly created shares as well ?

```suggestion
  function setVaultForToken(IERC20 token, IERC4626 vault, uint minAssetsOut, uint minNewShares) public virtual onlyAdmin {
```

src/strategies/utils/TransferLib2.sol
maxencerb, line 1: I don't know if this is from OZ, but a credit to OZ as well as specifying that OZ is under MIT license would be nice

src/strategies/offer_maker/market_making/kandel/ERC4626KandelSeeder.sol
```diff
@@ -58,9 +58,7 @@ contract ERC4626KandelSeeder is AbstractKandelSeeder {
     router.bind(address(kandel));
     // Set the Kandel as router admin
     router.setAdmin(address(kandel));
-    emit NewERC4626Kandel(
-      address(kandel), olKeyBaseQuote.hash(), olKeyBaseQuote.flipped().hash(), address(kandel), owner
-    );
+    emit NewERC4626Kandel(owner, olKeyBaseQuote.hash(), olKeyBaseQuote.flipped().hash(), address(kandel), owner);
```
maxencerb: Technically, it would be this:

```suggestion
    emit NewERC4626Kandel(owner, olKeyBaseQuote.hash(), olKeyBaseQuote.flipped().hash(), address(kandel), address(kandel));
```

src/strategies/offer_maker/market_making/kandel/abstract/CoreKandel.sol
maxencerb, line 1: keep the original transferlib here

src/strategies/routers/integrations/AavePooledRouter.sol
maxencerb, line 1: keep original transferlib

maxencerb (changes requested):
Keep the original TransferLib in the non audited contracts, this way there is no retro compat problems when we need to verify previously deployed contracts.

src/strategies/utils/TransferLib2.sol
Solthodox: Yes sure, its adapted from OZ

src/strategies/routers/integrations/ERC4626Router.sol
Solthodox: Roger that

maxencerb (approved):

PR #300, opened 2025-06-25
Feat/erc4626 routing logic

Solthodox:


src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+  IERC4626 public immutable VAULT;
+
+  /// @notice The underlying asset of the vault
+  IERC20 public immutable ASSET;
+
+  /// @notice Emitted when a deposit operation occurs
+  event VaultDeposit(uint assets, uint shares);
+
+  /// @notice Emitted when a withdrawal operation occurs
+  event VaultWithdraw(uint assets, uint shares);
```
maxencerb: I don't think we'll need events. It will increase the gas a lot, especially on megaeth

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+  /// @inheritdoc AbstractRoutingLogic
+  /// @notice Pulls tokens by withdrawing from the ERC4626 vault
+  /// @param token The token to pull (must match vault's asset)
+  /// @param fundOwner Not used - this contract owns the shares
+  /// @param amount The amount of underlying assets to withdraw
+  /// @param strict If true, must withdraw exact amount; if false, withdraw available balance
+  function pullLogic(IERC20 token, address fundOwner, uint amount, bool strict) external override returns (uint pulled) {
+    require(address(token) == address(ASSET), "SimpleERC4626Logic/InvalidToken");
+
+    if (withdrawAmount == 0) {
```
maxencerb: Does it compile ? I don't know where `withdrawAmount ` comes from.

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+      return 0;
+    }
+
+    // Withdraw exact amount
+    uint sharesUsed = VAULT.withdraw(withdrawAmount, msg.sender, address(this));
```
maxencerb: This contract does not hold any funds. This contract supposedly has allowance on the erc4626 tokens of the user. And it must transfer the funds from the user to the msg.sender by the end of the call. So modify like this:

```suggestion
    uint sharesUsed = VAULT.withdraw(amount, msg.sender, fundOwner);
```

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+
+    emit VaultWithdraw(withdrawAmount, sharesUsed);
```
maxencerb: remove the event

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+    // Transfer underlying assets from maker contract to this contract
+    require(
+      TransferLib.transferTokenFrom(token, msg.sender, address(this), amount), "SimpleERC4626Logic/AssetTransferFailed"
+    );
+
+    // Approve vault to spend the assets
+    require(TransferLib.approveToken(ASSET, address(VAULT), amount), "SimpleERC4626Logic/ApprovalFailed");
+
+    // Deposit into vault - this contract receives the shares
+    uint sharesReceived = VAULT.deposit(amount, address(this));
```
maxencerb: The receiver should be the fundOwner instead

```suggestion
    uint sharesReceived = VAULT.deposit(amount, fundOwner);
```

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+
+    emit VaultDeposit(amount, sharesReceived);
```
maxencerb: Remove the event

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+  /// @inheritdoc AbstractRoutingLogic
+  /// @notice Returns the underlying asset balance available from this contract's vault shares
+  /// @param token The token to check (must match vault's asset)
+  /// @param fundOwner Not used - this contract owns the shares
+  /// @return balance The amount of underlying assets this contract can withdraw
+  function balanceLogic(IERC20 token, address fundOwner) external view override returns (uint balance) {
+    require(address(token) == address(ASSET), "SimpleERC4626Logic/InvalidToken");
+
+    // Convert shares to underlying assets
+    balance = VAULT.convertToAssets(sharesBalance());
```
maxencerb: You should use `previewWithdraw` which is inclusive of the fees. And the holder of the funds is the `fundOwner` not the contract. Also this would not compile since you are calling an external function. You can remove other functions.

```suggestion
    balance = VAULT.previewWithdraw(VAULT.balanceOf(fundOwner));
```

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,116 @@
...
+    return VAULT.previewWithdraw(assets);
+  }
+
+  /// @notice Get vault information
+  /// @return vault The vault address
+  /// @return asset The underlying asset address
+  /// @return totalAssets The total assets in the vault
+  function getVaultInfo() external view returns (address vault, address asset, uint totalAssets) {
+    return (address(VAULT), address(ASSET), VAULT.totalAssets());
+  }
```
maxencerb: remove these functions

maxencerb (changes requested):

src/strategies/routing_logic/SimpleERC4626Logic.sol
Solthodox: this one should be `VAULT.maxWithdraw(fundOwner)`

src/strategies/routing_logic/SimpleERC4626Logic.sol
```diff
@@ -0,0 +1,79 @@
...
+
+    // Transfer underlying assets from maker contract to this contract
+    require(
+      TransferLib.transferTokenFrom(token, msg.sender, address(this), amount), "SimpleERC4626Logic/AssetTransferFailed"
+    );
+
+    // Approve vault to spend the assets
+    require(TransferLib.approveToken(ASSET, address(VAULT), amount), "SimpleERC4626Logic/ApprovalFailed");
+
+    // Deposit into vault - this contract receives the shares
```
maxencerb:
```suggestion
    // Deposit into vault - fund owner receives the shares
```

maxencerb (changes requested):
Just this comment, otherwise LGTM.

maxencerb (approved):

